PR #2, opened 2017-01-19
Improve comments for radio-group

robdodson:
Also fixup little errors here and there. @surma PTAL


elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
-// `RadioButton` is the radio button element itself. Maybe surprisingly,
-// there’s almost no logic on this element. 
-class RadioButton extends HTMLElement {
-  constructor() {
-    super();
-  }
+/**
+ * A `RadioGroup` is a set checkable buttons, where only one button may be
```
notwaldorf: typo: "set of checkable"

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+    /**
+     * The `RadioButton` sets its initial ARIA attributes when it's attached to
+     * the DOM. The surrounding RadioGroup handles dynamic changes to the ARIA
+     * attributes. The RadioButton should always set a `role` of `radio`, and
+     * should check to see if its `tabindex` and `aria-checked` values have been
+     * set by the user. Otherwise, it should use default values.
+     */
+    connectedCallback() {
+      this.setAttribute('role', 'radio');
+      this.setAttribute('tabindex', this.getAttribute('tabindex') || -1);
```
notwaldorf: I would maybe add a comment as to why you're setting this attribute at all (-1 vs no `tabindex`).wdyt?

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+     * already set to `tabindex=0`. This makes the first `RadioButton` focusable.
+     * This assumes `RadioButton` children are already in the DOM and their
+     * definition has already been loaded. For a more robust implementation you
+     * might consider using a Mutation Observer to detect if children are
+     * present yet. The `RadioGroup also adds listeners for keyboard and mouse
+     * events. 
+     */
+    connectedCallback() {
+      this.setAttribute('role', 'radiogroup');
+      let hasCheckedChild = this.querySelector('[role="radio"][tabindex="0"]');
```
surma: Theoretically the markup could contain multiple `radio-button` with `tabindex="0"`. Did you ignore that on purpose?

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
-    // but we don’t care. Lol.
-    this.querySelector('[role="radio"]').setAttribute('tabindex', 0);
+    /**
+     * The `RadioGroup` sets its ARIA role to `radiogroup` and sets the
+     * `tabindex` on its first `RadioButton` child to 0 if no other child is
+     * already set to `tabindex=0`. This makes the first `RadioButton` focusable.
+     * This assumes `RadioButton` children are already in the DOM and their
+     * definition has already been loaded. For a more robust implementation you
+     * might consider using a Mutation Observer to detect if children are
+     * present yet. The `RadioGroup also adds listeners for keyboard and mouse
```
surma: Missing \`

surma: ... I should probably upload the builds for PRs from CI to somewhere for review

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+     * the previous `RadioButton`. If the down or right keys were pressed,
+     * select the next `RadioButton`.
+     */
+    _onKeyDown(e) {
+      switch (e.keyCode) {
+        case KEYCODE.UP:
+        case KEYCODE.LEFT:
           e.preventDefault();
-          _this.setCheckedToPrevButton();
+          this._setCheckedToPrevButton();
```
surma: Considering you don’t `bind()` on `addEventListener`, can we rely on `this`?

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+    }
 
-  get firstRadioButton() {
-    return Array.from(this.querySelectorAll('[role="radio"]')).shift();
-  }
+    /**
+     * A getter for the first `RadioButton` child.
+     */
+    get firstRadioButton() {
+      return Array.from(this.querySelectorAll('[role="radio"]')).shift();
```
surma: _Arguably_ a little simpler (and memory-conservative?) to use `[role="radio"]:first-of-type`. WDYT?

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+    }
 
-  // Return the radio button that is next to the given one.
-  _prevRadioButton(node) {
-    let prev = node.previousElementSibling;
+    /**
+     * A getter for the last `RadioButton` child.
+     */
+    get lastRadioButton() {
+      return Array.from(this.querySelectorAll('[role="radio"]')).pop();
```
surma: Same here with `[role="radio"]:last-of-type`.

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,233 @@
...
+    }
 
-    while (prev) {
-      if (prev.getAttribute('role') === 'radio') {
-        return prev;
+    /**
+     * Return the previous sibling of the node that was passed in. The
```
surma: The comments on the following functions are kinda missing a first line that explains what the function does on a higher level. For example for this one I’d expect something like “Returns the radio button coming before the one passed as an argument”.

elements/radio-group/radio-group.js
robdodson: I guess i don't have a great reason except for consistency (eventually it may get set to -1 as part of the checked/unchecked process). I could remove this line entirely if you think that might be preferable.

elements/radio-group/radio-group.js
robdodson: Good question. The thought has occurred to me but I chose to ignore it for simplicity. I just did a demo to see what the native radio group does and it looks like if more than one child is checked, it styles whichever one is latest in the DOM. But it leaves `checked` on all of them.

Is there an approach you would prefer?

elements/radio-group/radio-group.js
robdodson: yes, we're listening for the keydown/click events on the `radio-group` itself, so the instance handling the event is always an instance of `radio-group`.

robdodson: if we were adding event listeners to our children then we would need to use `bind()`

elements/radio-group/radio-group.js
robdodson: sgtm

elements/radio-group/radio-group.js
surma: It’s up to you. [For the tab-panel](https://github.com/GoogleChrome/dash-elements/blob/master/elements/tab-panel/tab-panel.js#L61-L66) I chose to find the first tab that has `aria-selected` or – if none if them have – first tab in the order and call my `selectTab`, which unselects everything as a first step.

elements/radio-group/radio-group.js
surma: I was convinced that `this` would be `window`. TIL

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,241 @@
...
-      btn.setAttribute('aria-checked', 'false');
-      btn.tabIndex = -1;
+    /**
+     * Any user action (a keypress or mouse click) eventually funnels down to
+     * this method which ensures that only the passed in element is checked.
+     * Uncheck _all_ `RadioButton` children. Then set the `RadioButton` that was
+     * passed in to `aria-checked=true`. Also make it focusable with
+     * `tabIndex=0` and call its `focus()` method.
+     */
+    _setChecked(node) {
```
surma: Sorry, didn’t see this earlier – but maybe make this method public? Or is there another JS API to set the checked radio that I am missing?

elements/radio-group/radio-group.js
```diff
@@ -1,151 +1,241 @@
...
+    }
 
-  get firstRadioButton() {
-    return Array.from(this.querySelectorAll('[role="radio"]')).shift();
-  }
+    /**
+     * A getter for the first `RadioButton` child.
+     */
+    get firstRadioButton() {
+      return this.querySelectorAll('[role="radio"]:first-of-type');
```
surma: I don’t think you want `querySelectorAll` here, do you?

surma:
Also, please rebase this branch against `master` so we get the auto-build and the linting :)

elements/radio-group/radio-group.js
notwaldorf: Maybe leave a comment then you're doing it for consistency, but if you're worried about performance, don't do it? Although I think `tabindex` is pretty cheap

elements/radio-group/radio-group.js
robdodson: This is one of my favorite parts of custom elements. Because the DOM element is an actual instance of this class, `this` will refer to the class object. Way better than passing DOM elements into a constructor to wrap them.

elements/radio-group/radio-group.js
robdodson: nope, fixed

elements/radio-group/radio-group.js
robdodson: No method is provided. I would actually prefer to not provide a method but (if we have to) to instead only expose a property. But I don't think the native radio group provides any mechanism for doing this, other than setting `checked` yourself on an input[type=radio]

elements/radio-group/radio-group.js
robdodson: Because `selectTab` calls `focus()` at the end, I don't think this is something that should be done at startup. Otherwise the tab panel will steal focus before the user has interacted with it.

I'm thinking of factoring my `setChecked` method into three parts: uncheck everything, check the passed in node, and focus the passed in node. That way on init I can just call the first two steps but ignore the last one. wdyt?

robdodson:
@surma I've rebased against master. Also I named the actual elements `dash-*`. And I factored `_setChecked` into smaller functions so I can call them during initialization without stealing focus.

Not sure why, but my first block comment is appearing next to the code, instead of at the top of the doc like it does in tab-panel.js

surma:
Same as with Monica’s PR, I rebased against master so the new linting rules are used and I fixed the linter errors (mostly whitespace) and I renamed the files to `dash-*`.

[Preview](https://dash-elements.surma.link/f0eb92afdf8563341416fb0b0a2540a78d3e6409/dash-radio-group.html) is now available.

Are you going to add tests before we merge or do you want to do that separately?

elements/radio-group/radio-group.js
surma: Understood.

Do we want to handle the case when the `aria-selected` attribute is set via JS? [It works on the native element](https://jsbin.com/qevezer/1/edit?js,output).

robdodson:
I'll do it separate just because I figure it'll take a little while and I don't want to fall too far behind and get lost in rebase land :D

surma:
OFF WE GOOOOO 🚂

surma:
Merged in 115fdf015e37888bb5501dfb0200b2aecc260c0f

PR #7, opened 2017-01-25
Implement a dash-tooltip

notwaldorf:
I haven't added any tests yet because I can't tell if this is complete garbage. Any and all feedback is welcome.

i have no idea what im doing dog

/cc @robdodson @surma


elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,80 @@
...
+    // there's no tooltip to show.
+    this._target.removeEventListener('focus', this._show);
+    this._target.removeEventListener('blur', this._hide);
+    this._target.removeEventListener('mouseenter', this._show);
+    this._target.removeEventListener('mouseleave', this._hide);
+    this._target = null;
+  }
+
+  _show() {
+    // If the tooltip is hidden, show after a delay.
```
robdodson: Is there supposed to be a delay here?

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,80 @@
...
+    if (this.getAttribute('aria-hidden')) {
```
robdodson: I don't think this will work because `if ("false")` is truthy. I take it that this check is to prevent spamming `_show()` if the element is already set to `aria-hidden=false`, right?

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,37 @@
...
+  @keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+  }
+</style>
+
+<div class="text">
+   <label id="lb1" for="name">Your name:</label>
+   <input id="name" aria-labelledby="lb1" aria-describedby="tp1"/>
+   <dash-tooltip tabindex="0" id="tp1">Ideally your name is Batman</dash-tooltip>
```
robdodson: Are you setting `tabindex=0` here to assert that it gets set to `-1` when the element connects?

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,37 @@
...
+   <br>
+   <label id="lb2" for="noms">Favourite type of cheese: </label>
```
robdodson: No element with `id=noms`

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,37 @@
...
+
+  @keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+  }
+</style>
+
+<div class="text">
+   <label id="lb1" for="name">Your name:</label>
+   <input id="name" aria-labelledby="lb1" aria-describedby="tp1"/>
```
robdodson: Why're you using `aria-labelledby` if you are also using the `for` attribute of `<label>`?

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,37 @@
...
+   <dash-tooltip tabindex="0" id="tp1">Ideally your name is Batman</dash-tooltip>
+   <br>
+   <label id="lb2" for="noms">Favourite type of cheese: </label>
+   <input id="cheese" aria-labelledby="lb2" aria-describedby="tp2"/>
```
robdodson: Why're you using `aria-labelledby` if you are also using the `for` attribute of `<label>`?

surma:
I dared to do a rebase as I changed the linter so your tests pass (requiring block comments on evert function was a stupid idea).

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,37 @@
...
+  dash-tooltip {
+    margin-left: 8px;
+    padding: 4px;
+    background-color: #616161;
+    color: #fff;
+    border-radius: 3px;
+    opacity: 0.9;
+
+    /* Optionally, you can animate the tooltip. */
+    -webkit-animation: fadein 1s;
```
surma: I’d vote for removing the prefix’d properties. It’s more visual noise than anything else imo.

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,80 @@
...
+   * removed from the DOM. Theoretically, `connectedCallback` could be called
+   * multiple times when an element gets moved around.
+   */
+  disconnectedCallback() {
+    if (!this._target)
+      return;
+
+    // Remove the existing listeners, so that they don't trigger even though
+    // there's no tooltip to show.
+    this._target.removeEventListener('focus', this._show);
```
surma: I don’t think this `removeEventListener` does anything as you are binding to a different function above (because `.bind(this)`).

The @paullewis approach would be to add the following in the constructor:

```js
constructor() {
  //...
  this._show = this._show.bind(this);
  this._hide = this._hide.bind(this);
  // ...
}
```

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,80 @@
...
+    this._target.removeEventListener('blur', this._hide);
+    this._target.removeEventListener('mouseenter', this._show);
+    this._target.removeEventListener('mouseleave', this._hide);
+    this._target = null;
+  }
+
+  _show() {
+    // If the tooltip is hidden, show after a delay.
+    if (this.getAttribute('aria-hidden')) {
+      this.setAttribute('aria-hidden', false);
```
surma: Conversely, shouldn’t this be `"false"`?

robdodson: I think either is fine. If I recall correctly it'll translate `false` to `"false"`

elements/dash-tooltip/dash-tooltip.js
notwaldorf: Yes, but I didn't implement it because I wasn't sure if this is a thing the tooltip should do, or can be done as part of styling with the animation

elements/dash-tooltip/demo.html
notwaldorf: Oh shit, this is leftover from testing. Ignore!

notwaldorf:
@surma the `attributeChangedCallback` won't help, I don't think. It's not the tooltip's `aria-labelledby` that matters, it's _any_ element's (that points to the tooltip) so ughhhh. Listening to all the elements in the dom to see if they point to me sounds bad.

elements/dash-tooltip/dash-tooltip.js
notwaldorf: ✅

surma:
@notwaldorf Ah you are right. No, don’t listen to all the things. Maybe add a comment about then? Just describe the rationale.

Re Tests: Well, in case you hadn’t seen my [mind-blowing tests](https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.test.js)... fair point. I will whip up a PR for `dash-tab-panel` so we can discuss the approach to testing

surma:
LGTM. @robdodson @notwaldorf Merge at will :)

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+class DashTooltip extends HTMLElement {
+  /**
+  * The constructor does work that needs to be executed _exactly_ once.
+  */
+  constructor() {
+    super();
+
+    this._show = this._show.bind(this);
```
robdodson: Might comment why you're doing this

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,85 @@
...
+    this._hide = this._hide.bind(this);
+  }
+
+  /**
+   * `connectedCallback` is a life cycle callback defined by CustomElements v1
```
robdodson: I think lifecycle is one word

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,85 @@
...
+   * that gets called when an instance of the element gets inserted into the
+   * DOM. Theoretically, `connectedCallback` could be called multiple
+   * times when an element gets moved around.
+   */
+  connectedCallback() {
+    // A tooltip should always set its `role` to `tooltip`
+    this.setAttribute('role', 'tooltip');
+
+    // Tooltips cannot be focused themselves.
+    this.setAttribute('tabindex', -1);
```
robdodson: @surma @notwaldorf just curious, since `tabIndex` can also be set as a property, do we want to standardize on one approach?

elements/dash-tooltip/dash-tooltip.js
```diff
@@ -0,0 +1,85 @@
...
+    // The tooltip needs to listen to focus/blur events from the target,
+    // as well as hover events over the target.
+    this._target.addEventListener('focus', this._show);
+    this._target.addEventListener('blur', this._hide);
+    this._target.addEventListener('mouseenter', this._show);
+    this._target.addEventListener('mouseleave', this._hide);
+  }
+
+  /**
+   * `disconnectedCallback` is a life cycle callback defined by
```
robdodson: lifecycle

elements/dash-tooltip/dash-tooltip.js
robdodson: If it's not being implemented then the comment should be changed. If you handle it in CSS I think you could move the comment over there.

elements/dash-tooltip/dash-tooltip.js
notwaldorf: That's a good question! I'm ok with either

elements/dash-tooltip/dash-tooltip.js
notwaldorf: never copy comments from a german

elements/dash-tooltip/dash-tooltip.js
surma: Personally I avoid `setAttribute` whenever I can.

surma: ... having said that, I didn’t know that `tabIndex` is a property. DYK if it is wide supported in our browser range? If so I’d prefer it over `setAttribute`

notwaldorf:
@robdodson 👀 ✌️ 👀 ❓

elements/dash-tooltip/demo.html
```diff
@@ -0,0 +1,39 @@
...
+  dash-tooltip {
+    margin-left: 8px;
+    padding: 4px;
+    background-color: #616161;
+    color: #fff;
+    border-radius: 3px;
+    opacity: 0;
+
+    /* Optionally, animate the tooltip. */
+    will-change: opacity;
```
robdodson: Suggest adding a note on the `will-change` property so devs know what it does
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

robdodson:
Fine work @notwaldorf 🥇

surma:
🎉🎉🎉🎉

PR #12, opened 2017-01-31
Testing Infrastructure

surma:
HOLY SMOKES BATMAN 🦇

It was a journey but it works now 🎉 

I am not sure you want to review the actual test scaffolding code but _do_ take a look at the unit tests and the end-to-end tests and let me know if you think you can work with that.

The end-to-end tests are only run in Chrome for now as Firefox is being difficult about the keyboard events from Selenium. I’ll debug that at some point.

@robdodson @notwaldorf PTAL :)


run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+    && ['chrome'].includes(browser.getId());
+  if (!isChrome) return browser;
+  browser
+  .getSeleniumOptions()
+  .addArguments('--no-sandbox');
+  return browser;
+}
+
+async function main() {
+  // * Start a webserver to serve the docs so we can run the e2e tests on the
```
notwaldorf: Are these bullets or leftovers from a `/*` comment? Maybe uses dashes instead?

run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+  // demos.
+  // * require() all test suites.
+  // * Open all stable browsers and get their webdriver.
+  const [server, ...drivers] =
+    await Promise.all([
+      startServer(),
+      ...seleniumAssistant.getLocalBrowsers()
```
notwaldorf: omg `...`, this is the future

run-e2e-tests.js
surma: It’s supposed to be a list. Will turn into dashes

elements/dash-tab-panel/dash-tab-panel.e2etest.js
```diff
@@ -0,0 +1,28 @@
...
+        Key.TAB,
+        `return document.activeElement.getAttribute('aria-role') === 'tab';`
+      );
+    expect(found).to.equal(true);
+
+    await this.driver
+      .executeScript(`window.firstTab = document.activeElement;`);
+    await this.driver.actions().sendKeys(Key.ARROW_RIGHT).perform();
+    const rightFocus = await this.driver.executeScript(`
+      return window.firstTab.nextElementSibling === document.activeElement;
```
robdodson: What do you think about looking for the next element where `aria-role === 'tab'` instead of just looking for the next focusable sibling?

elements/dash-tab-panel/dash-tab-panel.unittest.js
```diff
@@ -0,0 +1,60 @@
...
+    })
+  );
+
+  it('should switch visibility when calling `selectTab()`',
+    dashElements.elementTest('dash-tab-panel', function() {
+      const selectedTab = this.tabs.find(t => t.classList.contains('selected'));
+      const selectedPanel =
+        this.panels.find(p =>
+          p.id === selectedTab.getAttribute('aria-controls'));
+      const otherTab = this.tabs.find(p => p !== selectedTab);
```
robdodson: Should this be `t =>` instead of `p =>`?

Personally I tend to prefer whole word variables for clarity. Since much of the work we're doing is meant for education purposes, it might help to do `tab =>` and `panel =>` especially for folks who are unfamiliar with `find/filter` and fat arrow syntax

elements/dash-tab-panel/dash-tab-panel.unittest.js
```diff
@@ -0,0 +1,60 @@
...
+  );
+
+  it('should have one selected tab initially',
+    dashElements.elementTest('dash-tab-panel', function() {
+      expect(this.tabs.filter(p => p.classList.contains('selected')))
+        .to.have.lengthOf(1);
+    })
+  );
+
+  it('should switch visibility when calling `selectTab()`',
```
robdodson: This is probably a longer discussion than one we want to have on this thread, but I was wondering if you might consider getting rid of the public `selectTab` method, and replacing it with a property. It would make the element more declarative and it would work better with libraries like React and Angular2 which favor a top down data flow model. I tried to explain my reasoning a bit here: http://robdodson.me/interoperable-custom-elements/#favoradeclarativeapioveranimperativeapi

run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+    elements.map(async element => {
+      const files = await fs.readdir(`elements/${element}/`);
+      if (files.some(f => /\.e2etest\.js/.test(f)))
+        return element;
+      else
+        return '';
+    })
+  );
+  filteredElements
+    .filter(name => !!name)
```
robdodson: Can you comment some of these lines to help folks grok what's going on

run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+    .then(_ => mocha);
+}
+
+function unsandboxChrome(browser) {
+  const isChrome = browser.getReleaseName() === 'stable'
+    && ['chrome'].includes(browser.getId());
+  if (!isChrome) return browser;
+  browser
+    .getSeleniumOptions()
+    .addArguments('--no-sandbox');
```
robdodson: What's the need for unsandboxing chrome? Is it to run ajax against the file system?

run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+      async (chain, driver) => {
+        await chain;
+        const mocha = await buildMocha();
+        mocha.suite.suites.forEach(s => {
+          s.ctx.driver = driver;
+          s.ctx.address = address;
+        });
+        return await runMocha(mocha).then(_ => {});
+      },
+      Promise.resolve()
```
robdodson: Can you comment what this function is doing?

I would imagine the return at line 106 would need to return a promise but my async/await skills are pretty newb still

run-e2e-tests.js
```diff
@@ -0,0 +1,120 @@
...
+    );
+}
+
+main()
+  .then(err => {
+    if (err) process.exit(1);
```
robdodson: I'm a little confused by the double handling of `err`. Can you explain what's going on here?

elements/dash-tab-panel/dash-tab-panel.unittest.js
```diff
@@ -0,0 +1,60 @@
...
+    this.tabs = Array.from(
+      this.container.querySelectorAll('[aria-role="tab"]')
+    );
+    this.panels = Array.from(
+      this.container.querySelectorAll('[aria-role="tabpanel"]')
+    );
+  });
+
+  it('should only have one visible panel initially',
+    dashElements.elementTest('dash-tab-panel', function() {
```
robdodson: Is it possible to use Mocha's `done` callback for async tests to move `elementTest` into the `beforeEach` callback? Then we'd only need to do it once I think.

elements/dash-tab-panel/dash-tab-panel.unittest.js
notwaldorf: Yes please

elements/dash-tab-panel/dash-tab-panel.unittest.js
surma: Mocha test functions are allowed to return promises now. But totally good point, will move that check to `beforeEach`.

elements/dash-tab-panel/dash-tab-panel.unittest.js
surma: Yes should at least be `t`, might actually go for longer names alltogether. Got a little lazy + frustrated with the tests after a while I guess ;)

run-e2e-tests.js
surma: Can do, I was actually meaning to ask you and @notwaldorf if we expect our tests to be “consumed”, too.

run-e2e-tests.js
surma: It just doesn’t run on Travis (or inside Docker) with sandboxing enabled.

run-e2e-tests.js
surma: It’s a workaround for now as node does not handling throwing inside async/await all that well just yet. So I am always resolving but with an error code.

elements/dash-tab-panel/dash-tab-panel.unittest.js
surma: That is an interesting discussion. It _feels_ more in line with native elements too. Not thinking about it very much, I am all for it. Will get back to you on this after BlinkOn.

run-e2e-tests.js
surma: It _is_ a promise because every `async` function returns a Promise by definition. I.e. if you return a plain value in an `async` function it will be wrapped in a Promise. I’ll add a few comments.

elements/dash-tab-panel/dash-tab-panel.unittest.js
surma: Made the function private for now.

surma:
@robdodson @notwaldorf PTAL :)

robdodson:
allow

surma:
WHOOOP! 🎉 🎉 🎉

Issue #13, opened 2017-02-06
Review Tab Panel


surma:
My tab panel pretty much made it to the repo without any actual reviews. I’d appreciate if either of you would give it a look to make sure I’m not doing random crap.

robdodson:
## Big Stuff:

The element isn't currently using ARIA correctly. See the [Roles, States, and Properties section of the ARIA Authoring Practices Guide](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-16). Also, make sure to do a pass with something like VoiceOver to see if it's announcing things properly. Usually VoiceOver will announce the role which can be an indicator if things are working as expected.

- The `<dash-tab-panel>` should self apply `role=tablist`. It may be better to rename this outter element to `<dash-tabs>` or `<dash-tablist>` to match its ARIA role.
- Element and demo use `aria-role=` but it should be `role=`
- Currently you say `aria-controls="tab1"` but it could be better expressed as `aria-controls="panel1" since a `tab` is really thought of as controlling a `tabpanel`.
- Each element with role `tabpanel` should have an `aria-labelledby` referring to its associated `tab` element.

Compare the difference in VoiceOver output (shown in black box at the bottom) when correct ARIA is used

**Before**:

![image](https://cloud.githubusercontent.com/assets/1066253/22752876/ede1507a-edee-11e6-8779-34ba03c06d84.png)


**After**:

![image](https://cloud.githubusercontent.com/assets/1066253/22752851/d5905b10-edee-11e6-8711-2ed67a85a4d5.png)


## Smaller Stuff:

- https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.js#L7 You don't need `aria-hidden` if something is `display: none` because `display: none` removes it from the a11y tree. Also you might reword this to say something like "inactive panels are hidden with display none" since the next section talks about selection. Otherwise it may not be clear that the two concepts are related.
- This opening section initially confused me https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.js#L14-L21. I wasn't sure which element was going to be a tab or a panel or a tabpanel. I had to look at the demo to figure it out.
- I would remove this comment https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.js#L30-L31. It's not really for accessibility, it's just for general usability that the element must handle the keyboard and mouse.
[...]

PR #17, opened 2017-02-14
<dash-tabs> review

surma:
I implemented your review, @robdodson. Here’s a few follow-up Qs/clarifications:

> - Each element with role `tabpanel` should have an `aria-labelledby` referring to its associated `tab` element.

Should I put that in the expected markup or is that something the code should apply for the developer? (For now, I chose the former.)

> - https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.js#L7 You don't need `aria-hidden` if something is `display: none` because `display: none` removes it from the a11y tree. Also you might reword this to say something like "inactive panels are hidden with display none" since the next section talks about selection. Otherwise it may not be clear that the two concepts are related.

I actually did that on purpose so that the element works with other hiding mechanism (e.g animations). I added a comment to clarify. WDYT?

> - You may want to add those listeners in `connectedCallback` so you can remove them in `disconnectedCallback`. 

I don’t quite see why that’s necessary. I’ll add the event listeners exactly once. So seems like the `constructor` is the perfect spot. 

> Also, we should get consistent with our handler names. I used `onClick` and you're using `handleClick`, throwing it to @notwaldorf to be the tie breaking vote :)

Good call. Adjusted to your naming convention. @notwaldorf: feel free to disagree.

> - Style nit: https://github.com/GoogleChrome/dash-elements/blob/master/elements/dash-tab-panel/dash-tab-panel.js#L47 I usually do `if (!tabs.length)`. @notwaldorf do you prefer one approach to the other?

No strong opinions, but I prefer not relying on truthiness/falsiness. 

> I'd be curious to see if it's possible to build the entire structure out of custom elements, something > like:
> 
> ```
> <dash-tabs>
>   <dash-tab></dash-tab>
>   <dash-tabpanel></dash-tabpanel>
> </dash-tabs>
> ```
>
> You could do some (possibly weird 😄 ) things for PE like give the tab a `role=heading` explicitly in > markup, but then that gets swapped out with `role=tab` when the element upgrades. The more I > think about this the more I kinda like this idea. It seems to dovetail well into the PE use case because it's saying "if JS doesn't load, this thing's role is something more basic, but if js _does_ load, it's new role is more advanced". Most importantly it gives the user a signal that the thing is now keyboard interactive.

I was thinking about if there’s a reason to do CEs for the inner elements. I settled for “no”. I was wrong :D I really like this as a way to do PE. Gonna punt on it for now, tho, and put it in a issue soI can get some more elements done.

@notwaldorf @robdodson PTAL :)


surma:
This closes #13

surma:
[Live preview](http://dash-elements.surma.link/19931ea400727da1e4800a37ef99938754f469a3/dash-tab-panel.html)

robdodson:
> Should I put that in the expected markup or is that something the code should apply for the developer? (For now, I chose the former.)

IMO, it'd be nice if the element just did this for you

> I actually did that on purpose so that the element works with other hiding mechanism (e.g animations). I added a comment to clarify. WDYT?

Suggest rewording to something like:

"The .hidden class might apply  display: none, which would remove the element from the accessibility tree anyway, but this way the element works when other hiding mechanism (like animation) are used."

> I don’t quite see why that’s necessary. I’ll add the event listeners exactly once. So seems like the constructor is the perfect spot.

I guess we need to check to make sure removing the element from the DOM cleans up the listeners and doesn't create a memory leak

> No strong opinions, but I prefer not relying on truthiness/falsiness.

@notwaldorf you make the call :)

surma:
> I guess we need to check to make sure removing the element from the DOM cleans up the listeners and doesn't create a memory leak.

It doesn’t! 🎉  ([Test JSBin](https://output.jsbin.com/nitiyif))
screenshot 2017-02-15 11 36 01

surma:
@robdodson Another question regarding `aria-labelledby`: If the tab doesn’t have an id I can use for the panel’s `aria-labelledby`, I currently ignore that. WDYT about a `console.warn`?

robdodson:
I think the memory leak test has a different outcome if you listen for a function that's on the prototype instead of creating an anonymous function. If the function is on the prototype it seems like it doesn't leak. But probably safest to encourage folks to clean up after themselves.

> Another question regarding aria-labelledby: If the tab doesn’t have an id I can use for the panel’s aria-labelledby, I currently ignore that. WDYT about a console.warn?

I guess a warn is ok. Another option, though maybe a bit involved, would be to check if the tab has an id, if not, auto-generate one. There's a group of folks kicking around the idea of writing accessibility mixins for custom elements and that's what they're doing (https://github.com/elix/rfcs/pull/4/files#diff-6f5adfb27ba10dc39a17b57a6107619aR173). But to keep things simple we could also just add a comment and be like "Hey if the thing doesn't have an id you could consider generating one"

surma:
If I [add a function on the prototype as the listener](https://output.jsbin.com/nitiyif/4/), the count stays constant o_O

screenshot 2017-02-16 13 11 17

surma:
> But probably safest to encourage folks to clean up after themselves.

This is actually a really good argument. Even though it might not be strictly necessary, it’s a good habit to advocate for. Will fix.

surma:
Implemented the last nits. I _think_ this is good to merge, PTAF(inal)L

robdodson:
What are you thoughts on ditching the `.hidden` and `.selected` classes in favor of using attr selectors of `[aria-hidden="true"]` and `[aria-selected="true"]`. The reason I suggest this is because it creates a kind of symbiotic relationship between the CSS and ARIA and let's you visualize if the states are being applied correctly

surma:
That is a good one. I vaguely remember some CSS folks being against it, but I don’t remember any good arguments, so as long as we don’t have those I am in favor.

robdodson:
I think technically attribute selectors are slower but I don't think we're in a situation where CSS performance is a bottleneck.

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -11,167 +11,216 @@
...
+      // element to the last child position.
+      tabs.forEach(tab => this.appendChild(tab));
+      panels.forEach(panel => this.appendChild(panel));
+
+      // Give each panel a `aria-labelledby` attribute that refers to the tab
+      // that controls it.
+      tabs.forEach(tab => {
+        const panelId = tab.getAttribute('aria-controls');
+        const panel = panels.find(panel => panel.id === panelId);
+        if (tab.id)
```
robdodson: feature request: we should add a js style linter thingy to this project so we have consistent braces. I always do them but am happy to not do them.

surma: Okay, I’ll add braces and add a rule to the linter

surma: On second thought, what about L136? Or L157? Feels much more noisy. Maybe allow skipping them for one-liners after all?

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -11,167 +11,216 @@
...
+      // their two groups. First all the tabs, then all the panels.
+      // Calling `appendChild` on an already inserted element _moves_ the
+      // element to the last child position.
+      tabs.forEach(tab => this.appendChild(tab));
+      panels.forEach(panel => this.appendChild(panel));
+
+      // Give each panel a `aria-labelledby` attribute that refers to the tab
+      // that controls it.
+      tabs.forEach(tab => {
+        const panelId = tab.getAttribute('aria-controls');
```
robdodson: This section is a little tricky because it assumes the user did write out some attributes (aria-controls) but didn't write out others (the id for the tab). What's the failsafe is `aria-controls` doesn't exist?

Thinking about this a bit more, I really feel like asking the user of the element to remember to put all the right aria-* or other required attributes on something is exposing an implementation detail. Hermetically sealing everything up into a `<dash-tab>` and `<dash-tabpanel>` would remove the need for a lot of error checking because you can assume that the right attributes will show up.

elements/dash-tab-panel/dash-tab-panel.js
robdodson: I'm fine skipping for one liners. Forgot we already had a linter :)

elements/dash-tab-panel/dash-tab-panel.js
surma: It’s a good argument for the inner CEs. The original reason why I didn’t do this because we are not necessarily going for developer _convenience_, but developer _education_. So I wonder if the layer of indirection through the inner CEs is going to “prevent” readers from understanding ARIA.

But combined with the PE argument above I should probably just do it™️ and write the inner CEs.

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+      // switching with arrow keys and Home/End.
+      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('click', this._onClick);
+
+      this.setAttribute('role', 'tablist');
+      // Acquire all tabs and panels inside the element
+      const tabs = Array.from(this.querySelectorAll('[role=tab]'));
+      const panels = Array.from(this.querySelectorAll('[role=tabpanel]'));
+      // If there are no tabs, there is no way to switch between panels. Abort.
+      if (!tabs.length) return;
```
robdodson: On second thought, I think your approach of not relying on falsiness is probably better. It's also more explicit which is good for teaching stuff. I vote we go back to !== 0

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+            `but needs one for ARIA.`);
+      });
+
+      // The element checks if any of the tabs have been marked as selected. If
+      // not, the first tab is now selected.
+      const selectedTab =
+        tabs.find(tab =>
+          tab.getAttribute('aria-selected') === 'true') || tabs[0];
+
+      // Next, we switch to the selected tab. `selectTab` takes care of marking
```
robdodson: I think we're avoiding the use of the word "We" in docs, right?

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
-   * `_handleClick` handles clicks inside the tab panel.
-   */
-  _handleClick(event) {
-    // If the click was not targeted on a tab element itself,
-    // it was a click inside the a panel or on empty space. Nothing to do.
-    if (!event.target.getAttribute('aria-role', 'tab')) return;
-    // If it was on a tab element, though, select that tab.
-    this._selectTab(event.target);
+    /**
+     * `selectTab` marks the given tab as selected.
```
robdodson: `_selectTab`

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+     * Additionally, it unhides the panel corresponding to the given tab.
+     */
+    _selectTab(newTab) {
+      // Deselect all tabs and hide all panels.
+      this.reset();
+
+      const panels = Array.from(this.querySelectorAll('[role=tabpanel]'));
+      // Get the panel that corresponds to the given tab.
+      const newPanelId = newTab.getAttribute('aria-controls');
+      const newPanel = panels.find(panel => panel.id === newPanelId);
```
robdodson: The above two lines are duplicated at 77,78. Might consider turning them into a helper function

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+     * `_onKeyDown` handles key presses inside the tab panel.
+     */
+    _onKeyDown(event) {
+      // The browser might have some native functionality bound to the arrow
+      // keys. Prevent the browser from taking any actions.
+      event.preventDefault();
+      // If the keypress did not originate from a tab element itself,
+      // it was a keypress inside the a panel or on empty space. Nothing to do.
+      if (event.target.getAttribute('role') !== 'tab') return;
+      // Don’t handle modifier shortcuts typically used by assistive devices.
```
robdodson: Suggest saying "assistive technology". A screen reader isn't really a device, it's just a piece of software

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+      if (event.altKey) return;
+
+      let newIdx;
+      let newTab;
+      const tabs = Array.from(this.querySelectorAll('[role=tab]'));
```
robdodson: Duplicated from 59. Suggest maybe defining getters for tabs and panels since they're accessed frequently

elements/dash-tab-panel/dash-tab-panel.js
```diff
@@ -4,174 +4,217 @@
...
+      // depending on the key that was pressed.
+      switch (event.keyCode) {
+        // When up or left is pressed, the previous tab is selected. If the
+        // first tab is reached, the element loops around to the last tab.
+        case KEYCODE.LEFT:
+        case KEYCODE.UP:
+          // The element uses `findIndex` to find the index of the currently
+          // selected element and subtracts one to get the index of the previous
+          // element.
+          newIdx =
```
robdodson: It might be nice to turn these into helper methods like `_prevTab` and `_nextTab`. Then you could say _selectNextTab or _selectPrevTab inside of these switch statements. That's what I did in the radio-group. That way we could start to zero in on some consistent patterns since I imagine the big keyboard switch/case will be pretty common in most of our elements

elements/dash-tab-panel/dash-tab-panel.js
surma: Dammit

elements/dash-tab-panel/dash-tab-panel.js
surma: What’s a good alternative?

surma: I mean, I am fine with “screen readers” but I thought that might not be broad enough.

elements/dash-tab-panel/dash-tab-panel.js
robdodson: Yeah I get that. My hope is we can show folks that they can use ARIA in a "modern" fashion where the element is smart and self applies it. That way folks don't feel like they have to spray their hand written markup with attributes which can be hard to remember.

A possible idea is we do a second pass in the future and really explain what the ARIA attributes are doing in the JS. So rather than just saying something like "foo() sets aria-checked to true" we could spend a bit of time in the comment diving into what it means for aria-checked to be true for assistive technology and the accessibility tree. I'll have to think on that a bit more :)

elements/dash-tab-panel/dash-tab-panel.js
robdodson: oh i meant say "assistive technology" in place of "assistive devices"

surma: My bad. Will fix.

surma:
PTAL! I added `<dash-tab>` and `<dash-tabpanel>`, renamed the main element to `<dash-tabs>` and did a few other things. Did a pass on the commentray as well.

@robdodson I renamed all my classes to `DashElementName` so that class name and tag name correspond, like @notwaldorf did in her code. I’d be in favor of updating the radiobox as well.

[Live version](http://dash-elements.surma.link/211ea11b8dbbcfce37a4fc36d26b17483fbe82ad/dash-tabs.html)

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+     * `_allPanels` returns all the panels in the tab panel. This function could
+     * memoize the result if the DOM queries ever become a performance issue.
+     * The downside of memoization is that dynamically added tabs and panels
+     * will not be handled.
+     *
+     * This is a method and not a getter, because a getter implies that it is
+     * cheap to read.
+     */
+    _allPanels() {
+      return Array.from(this.querySelectorAll('dash-tabpanel'));
```
robdodson: I noticed in Chrome at least you don't need `Array.from` here. NodeLists seem to be iterable and have `forEach`. Not sure if this is supported everywhere though

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+
+        // When right or down is pressed, it’s the next tab.
+        case KEYCODE.RIGHT:
+        case KEYCODE.DOWN:
+          newTab = this._nextTab();
+          break;
+
+        // The home button always selects the first tab.
+        case KEYCODE.HOME:
+          newTab = this._allTabs()[0];
```
robdodson: What do you think about adding a _firstTab and _lastTab getters? I think I noticed a handful of `[0]`s sprinkled around

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+  class DashTab extends HTMLElement {
+    constructor() {
+      super();
+    }
+
+    connectedCallback() {
+      // If this is executed, JavaScript is working and the element
+      // changes its role to `tab`.
+      this.setAttribute('role', 'tab');
+      if (!this.id) console.warn('All DashTabs must have an ID');
```
robdodson: I think you need to `return` here otherwise when you try to access `this.id` on 272 it'll be undefined

elements/dash-tabs/dash-tabs.js
surma: Yes, `NodeList` has `forEach` in Chrome, Safari and Firefox, but not `find`, `findIndex` or `filter`. So I’d rather convert to a proper array.

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+     */
+    connectedCallback() {
+      // The element needs to do some manual input event handling to allow
+      // switching with arrow keys and Home/End.
+      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('click', this._onClick);
+
+      this.setAttribute('role', 'tablist');
+      // Acquire all tabs and panels inside the element
+      const tabs = this._allTabs();
```
surma: I explained that decision in the comment of `_allTabs()`. WDYT?

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+    _allTabs() {
+      return Array.from(this.querySelectorAll('dash-tab'));
+    }
+
+    /**
+     * `_panelForTab` returns the panel that the given tab controls.
+     */
+    _panelForTab(tab) {
+      const panelId = tab.getAttribute('aria-controls');
+      return this._allPanels().find(panel => panel.id === panelId);
```
surma: That’s actually a good point since IDs are globally unique (and probably optimized to query).

elements/dash-tabs/dash-tabs.js
surma: I did this for the same reason I made `_allTabs()` a function not a setter, to not hide the cost of this operation.

elements/dash-tabs/dash-tabs.js
```diff
@@ -0,0 +1,293 @@
...
+      // First all the tabs, then all the panels.
+      // Calling `appendChild` on an already inserted element _moves_ the
+      // element to the last position.
+      tabs.forEach(tab => this.appendChild(tab));
+      panels.forEach(panel => this.appendChild(panel));
+
+      // Give each panel a `aria-labelledby` attribute that refers to the tab
+      // that controls it.
+      tabs.forEach(tab => {
+        if (tab.id)
```
surma: Yes, you mentioned that. We can easily add that to `<dash-tab>` and/or `<dash-tabpanel>`. You think I should do it in v1?

surma: I guess I am asking if we should spent LOCs on developer convenience. When the primary goal is for the source to be read, LOCs are arguably more expensive, but OTOH just doing something `` element.id = generated-tab-id-${Date.now()}` `` is not really that expensive in terms of cognitive load.

elements/dash-tabs/dash-tabs.js
robdodson: sgtm

elements/dash-tabs/dash-tabs.js
robdodson: I agree on the point about developer convenience. I was thinking of it more as a suggestion for building "modern" accessible components. It shifts the burden of applying the right ids onto the component author (where it belongs, IMO), instead of the component user. If our target audience is component authors, this might be a useful pattern to teach them.

elements/dash-tabs/dash-tabs.js
robdodson: OK, we can switch to function calls then.

So... what do you think about a _firstTab() and _lastTab() ?  😸

robdodson:
@surma thank you for putting up with my world's longest code review. you deserve a medal 🏅

elements/dash-tabs/dash-tabs.js
notwaldorf: My 2c: In `paper-input`, we literally have a count on the prototype and we keep adding 1 to it every time an instance is created and it's never been bad.  Or date is fine too. I don't know if I'd import a library that generates a random ID for me.

notwaldorf:
I am just gonna leave answers to all the things I've been summoned on, because I've been all-in working on the polymer site and ignored y'all.
- i prefer 'onEventName', and even more in particular, '_onClick', to pretend we have public/private methods.
- i am ok with truthiness! `if (!tabs.length)` is ok, though a bunch of times i will do specific checks for `null` or `undefined` because they come up a lot in custom elements (i've noticed). But like, tbh, this sounds like style and I wouldn't even mind if it was different across elements.

notwaldorf:
I'm rull excited about how nice this element is coming along btw!!

elements/dash-tabs/dash-tabs.js
surma: >  If our target audience is component authors, this might be a useful pattern to teach them.

Fair point. I’ll add it.

> I don't know if I'd import a library that generates a random ID for me.

No, I wouldn’t, either. Counter or time-based should be fine. I’ll whip something up :)

elements/dash-tabs/dash-tabs.js
surma: Eh, pretty indifferent about that one. But I guess in terms of readability it’s a win, so I’ll add it :)

surma:
> i am ok with truthiness! 

Imma call overruled here 😛 

PTAL! Specifically, at `DashTabs.connectedCallback()`, because i am now relying on `<dash-tab>` to generate IDs, I needed to introduce `customElements.whenDefined()` to wait for the tabs to be initialized. I think it could be a useful pattern (I think @robdodson had a similar problem in `DashRadioGroup`). (I can haz async/await?.

surma:
[Live preview](http://dash-elements.surma.link/31312e946446fe5aa36ab8e80c7c47f7aac1c182/dash-tab-panel.html)

robdodson:
OK last few points I swear :)

- It seems inconsistent for the tab to generate its own id but not the tabpanel. Feels like they should both generate ids.
- It feels like `aria-controls` could be applied by the CE when it boots up. Explicitly having a `heading` that `aria-controls` a group or region doesn't make much sense. Technically [it's allowed](https://www.w3.org/TR/wai-aria-1.1/#heading) but it seems confusing.
- For the PE use case, let's apply [`role=region`](https://www.w3.org/TR/wai-aria-1.1/#region) to the unupgraded tabpanel. Unfortunately `role=section` is abstract and we're not allowed to use it. `role=region` seems like the next best thing. I ran it by alice and she approved 👍. And it's pretty easy to change later if other folks in the community disagree.

surma:
> - It seems inconsistent for the tab to generate its own id but not the tabpanel. Feels like they should both generate ids.
> - It feels like `aria-controls` could be applied by the CE when it boots up. Explicitly having a `heading` that `aria-controls` a group or region doesn't make much sense. Technically [it's allowed](https://www.w3.org/TR/wai-aria-1.1/#heading) but it seems confusing.

My thought process behind that was that the developer has to link up tab and panel anyways. I could instead generate all the ids and rely on DOM order to figure out which tab belongs to which panel.

robdodson:
> I could instead generate all the ids and rely on DOM order to figure out which tab belongs to which panel.

That's kind of what I was thinking. ARIA should be an implementation detail not exposed to the component consumer. I ran this whole idea behind alice and she approved. My only nit (and we don't have to do this for this PR, just mentioning it) is that it would be a bit nicer to do something like:

```
<tab>
  Tab 1
  <tabpanel>
</tab>
```

Because then the relationship is easier to grok and it starts to work like `<label>`. For now we can just leave them as siblings.

robdodson:
awwww yissssss

giphy

surma:
🎉 🎉 🎉 🎉 🎉

notwaldorf:
Well actually, I had a couple more comments...

JK, this looks amazing! Well done fam. ❤️❤️❤️

PR #25, opened 2017-02-23
Minor markup fixes for a11y

surma:
[TIL](https://twitter.com/DasSurma/status/834713623851302912) that the `lang` attribute on `<html>` is important, so I fixed that.

I also did a pass with a screen reader on the current site and noticed that demo feels awkward. Tried to improve it, please double-check I’m not doing anything dumb here, @robdodson 😅 

The code section feels awkward, too, but I’m not sure how to address that.

### DIFF
--- a/site-resources/demo.tpl.html
+++ b/site-resources/demo.tpl.html
@@ -1,4 +1,5 @@
 <!doctype html>
+<html lang="en">
 <p>
   <a href="?nojs">Load without JavaScript</a>
   <a href="?">Load with JavaScript</a>
@@ -34,3 +35,4 @@
   window.addEventListener('message', sendSize);
   sendSize();
 </script>
+</html>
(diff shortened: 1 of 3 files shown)


robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
+1 for `lang` fixes.

I don't think we can use the `section` role as the ARIA spec recommends against directly using abstract roles (https://www.w3.org/TR/wai-aria-1.1/#abstract_roles). Let's go with `region` instead.

Also looks like we have a duplicate id on each page, for example:

`<ul class="literate demo" id="dash-tabs">` got called out by axe so somewhere else there's a `dash-tabs` id.

Also we have 200+ low contrast text warnings 😨

surma:
> Also we have 200+ low contrast text warnings 😨

Yeah, I am aware. I was thinking of addressing those when we have the logo and have to figure out a page design :)

robdodson:
> Yeah, I am aware. I was thinking of addressing those when we have the logo and have to figure out a page design :)

sgtm

PR #29, opened 2017-02-27
Add <dash-tree>

robdodson:
Since I'm about to disappear to CSUN for a week I wanted to get this out for code review. I still need to add more comments and tests but the implementation is all there. I noticed VoiceOver isn't announcing the element properly so i need to do some research there. I know activedescendant has some issues on mac :\

@surma @notwaldorf PTAL


robdodson:
oh also... the demo gets cut off because the iframe doesn't dynamically expand :\

surma:
> oh also... the demo gets cut off because the iframe doesn't dynamically expand :\

Yeah I ran into this with `<dash-accordion>`, too. I’ll look into it. Sadly, `ResizeObserver` is not a thing yet 😕

surma:
Also: Fix dem linter tests so we can have a preview ;)

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+
+      // There can be only one active item at a time.
+      // Find any previous active item and remove its active class.
+      const activeItem = this.querySelector('.active');
+      if (activeItem)
+        activeItem.classList.remove('active');
+      // If the element is expandable then it should have a <span> label
+      // as an immediate child. Put the `active` class on that child instead.
+      if (treeItem.hasAttribute('aria-expanded')) {
+        treeItem.firstElementChild.classList.add('active');
```
surma: Is there any way we can have the `.active` class be on `<dash-treeitem>` regardless of expandability?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+     * node, dispatch an event with a reference to the node. If this was
+     * a file picker, an application could listen for this event and open
+     * the file based on the item's name.
+     */
+    _selectTreeItem() {
+      // Set the item to aria-selected and remove aria-selected from
+      // any other nodes
+      const treeItem = this._currentTreeItem();
+      const treeItems = this.querySelectorAll('dash-treeitem');
+      for (let i = 0; i < treeItems.length; i++) {
```
surma: `forEach`?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+        treeItems[i].removeAttribute('aria-selected');
```
surma: Why not also remove `.selected` here?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+
+      if (treeItem.hasAttribute('aria-expanded')) {
+        if (isExpanded(treeItem)) {
+          this._collapseTreeItem();
+        } else {
+          this._expandTreeItem();
+        }
+        // Since this is an expandable element, set the selected class
+        // on its `<span>` label.
+        treeItem.firstElementChild.classList.add('selected');
```
surma: So this branch doesn’t send a `dash-tree-item-selected` event. Is that intentional?

elements/dash-tree/demo.html
```diff
@@ -0,0 +1,97 @@
+<style>
+  dash-tree {
+    font-family: 'Courier New', Courier, monospace;
+  }
+  dash-tree,
+  dash-treeitem {
```
surma: Apply this to `dash-treegroup` as well

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+  class DashTree extends HTMLElement {
+    constructor() {
+      super();
+    }
+
+    connectedCallback() {
+      // The element needs to do some manual input event handling to allow
+      // switching with arrow keys and Home/End.
+      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('click', this._onClick);
```
surma: Don’t see a `_onClick`

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+
+      this.setAttribute('role', 'treeitem');
+
+      // If the element doesn't already have an ID, generate one for it.
+      // This will make it easier to set the element as active using
+      // aria-activedescendant.
```
surma: backticks

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+            if (el.firstElementChild) {
+              findTreeitems(el, isVisible);
+              return;
+            }
+          }
+
+          if (isTreeItem(el)) {
+            treeItems.push(el);
+            el.visible = isVisible;
+            // If at any point we hit a parent with aria-expanded=false
```
surma: backticks

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      }
+
+      // The browser might have some native functionality bound to the arrow
+      // keys, home or end. The element calls `preventDefault` to prevent the
+      // browser from taking any actions.
+      event.preventDefault();
+    }
+
+    /**
+     * Return the current activedescendant if there is one. Otherwise,
```
surma: Either backticks or two words

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+        // on its `<span>` label.
+        treeItem.firstElementChild.classList.add('selected');
+        return;
+      }
+      treeItem.classList.add('selected');
+
+      // Dispatch a non-bubbling event containing a reference to the selected
+      // node. The reason to choose non-bubbling is explained in this post
+      // https://medium.com/dev-channel/custom-elements-that-work-anywhere-898e1dd2bc48#.w6ww4mgfc
+      // TODO: How do you deep link a Medium post???
```
surma: Heh. I should probably adhere to this rule in `<dash-accordion>`, too.

I wonder if this is one of the resources we link to in the intro. Should we try to do a paper-like reference system? I could augment the parser to auto-link the respective sites.

Then you’d write something like `[Events from custom elements should only rarely bubble][bubbling-events].`

WDYT?

surma:
I fixed the parser so now code doesn’t disappear anymore ;) I left the commit in this branch although it’s technically only related

surma:
If you rebase this branch against master you’ll have the iframe-auto-size feature :)

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      // This will make it easier to set the element as active using
+      // aria-activedescendant.
+      if (!this.id)
+        this.id = `dash-treeitem-generated-${dashTreeItemCounter++}`;
+
+      // If the element contains a child then it's a parent node
+      if (this.firstElementChild) {
+          // If the element is not explicitly already expanded by the user, then
+          // set it to closed.
+          if (!isExpanded(this))
```
robdodson: Do you mean adding an isNotExpanded method? Or do you mean converting isExpanded to being a method on the class instead of a helper?

surma: I think this would make sense as a method on the class instead of a free-floating helper :)

robdodson:
Btw I figured out the issue with VoiceOver. It was in quick nav mode. Jesse Beach showed me how to turn it off. I think it's VO keys plus left and right at the same time.

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+  class DashTreeItem extends HTMLElement {
+    constructor() {
+      super();
+    }
+
+    connectedCallback() {
+      // The element needs to do some manual input event handling to allow
+      // switching with arrow keys and Home/End.
+      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('click', this._onClick);
```
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      this.setAttribute('role', 'treeitem');
+
+      // If the element doesn't already have an ID, generate one for it.
+      // This will make it easier to set the element as active using
+      // aria-activedescendant.
+      if (!this.id)
+        this.id = `dash-treeitem-generated-${dashTreeItemCounter++}`;
+
+      // If the element contains a child then it's a parent node
+      if (this.firstElementChild) {
```
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: I see. Previously I had a few of these as class methods but because I use them in the recursive _findTreeItems function it got a little wonky having to keep rebinding to `this`. Let me see if it's a little simpler now that I only have this one method.

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      });
+    }
+
+    _allTreeItems() {
+      const treeItems = [];
+      // A recursive function that visits every child and builds a tree
+      // This produces the same results as calling querySelectorAll,
+      // but writing the tree ourselves lets us toggle which elements
+      // are visible.
+      function findTreeitems(node, isVisible) {
```
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+        let el = node.firstElementChild;
+        while (el) {
```
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+          if (!isTreeItem(el)) {
+            if (el.firstElementChild) {
+              findTreeitems(el, isVisible);
+              return;
+            }
+          }
+
+          if (isTreeItem(el)) {
+            treeItems.push(el);
+            el.visible = isVisible;
```
robdodson: yeah it's kind of a quasi private attribute. It makes it really easy to figure out if I can send focus to a child, but it's not something I'd ever want the user to set. @notwaldorf are there any examples of properties like this in Polymer land?

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      // browser from taking any actions.
+      event.preventDefault();
+    }
+
+    /**
+     * Return the current activedescendant if there is one. Otherwise,
+     * return null.
+     */
+    _currentTreeItem() {
+      const activeDescendant = this.getAttribute('aria-activeDescendant');
```
robdodson: oh that was a typo. fixed

elements/dash-tree/dash-tree.js
robdodson: nevermind, I figured out how to do this with a simpler tree walk. Basically as soon as I hit an item that's not visible I stop walking the tree in that direction. This simplifies going next/prev as well since invisible nodes aren't even in the tree.

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+    /**
+     * `_nextTreeItem` gets the treeitem that comes after the currently active
+     * one.
+     */
+    _focusNextTreeItem() {
+      const treeItems = this._allTreeItems();
+      const currentItem = this._currentTreeItem();
+
+      let next = undefined;
+      for (let i = treeItems.length - 1; i >= 0; i--) {
```
robdodson: yeah I changed how the tree was built and was able to simplify this.

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+      }
+      if (prev)
+        this._focusTreeItem(prev);
+    }
+
+    /**
+     * Focus the first `<dash-treeitem>` in the tree.
+     */
+    _focusFirstTreeItem() {
+      this._focusTreeItem(this.querySelector('dash-treeitem'));
```
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    /**
+     * Focus the last `<dash-treeitem>` in the tree that is focusable without
+     * opening a node.
+     */
+    _focusLastTreeItem() {
+      const treeItems = this._allTreeItems();
```
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: yeah that's what the line after the return does

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: probably should

elements/dash-tree/dash-tree.js
robdodson: Hm that could be useful but I don't want to add work to your plate :)

elements/dash-tree/demo.html
robdodson: fixed

robdodson:
OK did a lot of updating. @surma PTAL

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Define a custom element, `<dash-treeitem>`, and associate it with the
+   * `DashTreeItem` class.
+   */
+  window.customElements.define('dash-treeitem', DashTreeItem);
```
surma: Don’t punch me, but for consistency’s sake, shouldn’t it be `dash-tree-item` and `dash-tree-group`?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+        if (!isExpanded(this))
+          this.setAttribute('aria-expanded', false);
+
+        // Set the `aria-label` to the first bit of text. Otherwise the label
+        // for this element will be computed based on its text content plus
+        // the text content of all of its children. Making it so verbose as to
+        // be unusable. This first child should ideally be a `<span>` element,
+        // but it may also just be a line of text.
+        if (!this.hasAttribute('aria-label')) {
+          let label = this.querySelector('span');
```
surma: Didn’t think of this before because I hadn’t started the combobox yet, but would `<label>` instead of `<span>` be appropriate here?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+      // This will make it easier to set the element as active using
+      // `aria-activedescendant`.
+      if (!this.id)
+        this.id = `dash-treeitem-generated-${dashTreeItemCounter++}`;
+
+      // If the element contains a `<dash-treegroup>` then it's a parent node
+      if (this.querySelector('dash-treegroup')) {
+        // If the element is not explicitly already expanded by the user, then
+        // set it to closed.
+        if (!isExpanded(this))
```
surma: I still think this should be a method like `this._isExpanded()`

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+      // This produces similar results to calling querySelectorAll,
+      // but allows for filtering of the children based on whether or not
+      // their parent is currently visible.
+      function findTreeItems(node, isVisible) {
+        for (let el of node.children) {
+          // Ignore children like `<span>` and `<dash-treegroup>`.
+          // If the element has children, descend into them looking for
+          // more treeitems.
+          if (!isTreeItem(el)) {
+            if (el.firstElementChild) {
```
surma: Wanna combine these into one if like below?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+     * If an item is a child of a parent who is set to `aria-expanded=false`
+     * then it is considered invisible and is not added to the list.
+     */
+    _allTreeItems() {
+      const treeItems = [];
+      // A recursive function that visits every child and builds a list
+      // This produces similar results to calling querySelectorAll,
+      // but allows for filtering of the children based on whether or not
+      // their parent is currently visible.
+      function findTreeItems(node, isVisible) {
```
surma: Do you still need the `isVisible` parameter at all? You start with `true` and never change it.

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+     * Find the `DashTreeItem` associated with the element that was clicked.
+     * Focus the treeitem and make it the current selected item as well.
+     */
+    _onClick(event) {
+      let treeItem = undefined;
+      // A recursive function that will work its way upward until it finds
+      // the `DashTreeItem` associated with the event target. This allows
+      // clicking on a `<span>` or `DashTreeGroup` within a `DashTreeItem` and
+      // ensures the right element is always being focused/selected.
+      function findTreeItem(node) {
```
surma: Isn’t this basically a `while` loop?

elements/dash-tree/dash-tree.js
surma: Wait I don’t think I understand. 

`.active` is on `<dash-treeitem>`, except if it expandable, then it’s on the `<span>` inside `<dash-treeitem>`. Right? And if yes, can’t we have `.active` always be on `<dash-treeitem>` and adjust the styling accordingly?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,448 @@
...
+        }
+        treeItem.querySelector('span').classList.add('selected');
+      } else {
+        treeItem.classList.add('selected');
+      }
+
+      // Dispatch a non-bubbling event containing a reference to the selected
+      // node. The reason to choose non-bubbling is explained in this post
+      // https://medium.com/dev-channel/custom-elements-that-work-anywhere-898e1dd2bc48#.w6ww4mgfc
+      dispatchEvent(new CustomEvent('dash-tree-item-selected', {
```
surma: `this.dispatchEvent`?

surma:
Also, the click handler doesn’t deselect properly:

screenshot 2017-03-07 14 01 10

elements/dash-tree/dash-tree.js
robdodson: I wanted to do that but wasn't sure how to go about it. Maybe i'm having a total brainfart :)

.active sets the background color so doing:

```
.active,
.active > span {
  background: grey;
}
```

you'll end up with the entire treeitem having a grey background, even if it contains a span

elements/dash-tree/dash-tree.js
robdodson: yeah not sure to be honest. `<label>` only works with specific input elements, select and textarea. But it may be nicer to use it since it semantically conveys what we're going for. And hopefully a future version of Custom Elements will be able to work with label elements. I'll add it in and see how Alice feels :D

elements/dash-tree/dash-tree.js
surma: Since the `<span>` is semantically a label, I’d require a `<span>` on leafs _and_ inner nodes. The distinction of leafs and inner nodes happens with a contained `<dash-itemgroup>`. You can add the `<span>` automatically if it’s missing maybe. WDYT?

surma: ... also we need `:matches()` 😉

robdodson: It's reasonable to allow items to have their label computed by text content. That's pretty common. Think of `<li>`.

In this case the only reason we have a label is because if you don't apply `aria-label` to an element with children, then it will compute the label based on the text content of all the children. I think I'd prefer the js check versus adding additional labels to every treeitem. @alice do you have an opinion?

> ... also we need :matches()

Yeah Alice and I were discussing this a while back when I first wrote the element. I initially wanted to do all of this with CSS. I bet `:matches()` is expensive to compute though.

elements/dash-tree/dash-tree.js
robdodson: yeah that was left over, I think I can drop it

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/dash-tree.js
surma: I think performance was one of the major blockers on `:matches` :(

You are right re label. What do you think about this?

```css
.active:not([aria-expanded]), 
.active[aria-expanded] > span {
  /* ... */
}
```

elements/dash-tree/dash-tree.js
robdodson: yep fixed

elements/dash-tree/dash-tree.js
robdodson: I really prefer leaving it as a helper as it avoids needing to bind it in `findTreeItems`. All the code is wrapped in an IIFE so it's not a global

surma: Fair enough.

elements/dash-tree/dash-tree.js
robdodson: fixed

robdodson:
ok @surma I believe this is ready for another review

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
...
+        this._expandTreeItem();
+      }
+    }
+
+    /**
+     * If focus is on an open node, closes the node.
+     * If focus is on a child node that is also either an end node or a closed
+     * node, move focus to its parent node.
+     */
+    _collapseTreeItem() {
```
surma: I’d either call this `_collapseCurrentTreeItem()` or – and I prefer that – have this take a parameter.

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
...
+          item.classList.remove('selected');
+        });
+
+      const treeItem = this._currentTreeItem();
+      treeItem.setAttribute('aria-selected', 'true');
+      treeItem.classList.add('selected');
+
+      // Dispatch a non-bubbling event containing a reference to the selected
+      // node. The reason to choose non-bubbling is explained in this post
+      // https://medium.com/dev-channel/custom-elements-that-work-anywhere-898e1dd2bc48#.w6ww4mgfc
```
surma: maybe rater `[this post](... link ...)`? Looks better on [the page](http://dash-elements.surma.link/a6f34742f35b7aba2a838ccbea368c29964c8f9c/dash-tree.html).

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
+/**
+ * A `DashTree` presents a hierarchical list of children. Each node in the tree
+ * is represented as a `DashTreeItem` element. If a `DashTreeItem` contains only
+ * text it is considered an "end node". If a `DashTreeItem` contains children,
+ * it is considered a "parent node". If a `DashTreeItem` is a parent node, its
```
robdodson: I took the terminology from the ARIA authoring practices guide but I can change it to leaf/parent

robdodson: I went with "leaf" and "parent", wdyt?

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
...
+ * Parent nodes can be either collapsed or expanded to reveal their children.
+ * The state of the parent node is conveyed through the use of a `aria-expanded`
+ * attribute.
+ *
+ * Depending on the implementation, trees can support either single or
+ * multi selection. The `DashTree` element supports single selection, so
+ * there can only be one selected element at a time. The currently selected
+ * element is indicated by the `aria-selected` attribute.
+ *
+ * Unlike the `DashRadioGroup`, which uses roving tabindex to indicate which
```
robdodson: fixed

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
...
+      this.setAttribute('role', 'treeitem');
+
+      // If the element doesn't already have an `id`, generate one for it.
+      // This will make it easier to set the element as active using
+      // `aria-activedescendant`.
+      if (!this.id)
+        this.id = `dash-tree-item-generated-${dashTreeItemCounter++}`;
+
+      // If the element contains a `<dash-tree-group>` then it's a parent node
+      if (this.querySelector('dash-tree-group')) {
```
robdodson: added `isParentNode()`

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,436 @@
...
+      // A recursive function that visits every child and builds a list
+      // This produces similar results to calling querySelectorAll,
+      // but allows for filtering of the children based on whether or not
+      // their parent is currently expanded.
+      function findTreeItems(node) {
+        for (let el of node.children) {
+          // Ignore children like `<label>`, and `<dash-tree-group>`.
+          // If the element has children, descend into them looking for
+          // more treeitems.
+          if (!isTreeItem(el) && el.firstElementChild) {
```
robdodson: yeah I was able to go with your approach. Just needed to swap out the `return` with a `continue`. Otherwise it _seems_ to work as expected 👍

elements/dash-tree/dash-tree.js
robdodson: yes! I think I may parameterize all of these

robdodson: fixed

elements/dash-tree/dash-tree.js
robdodson: fixed

elements/dash-tree/demo.html
```diff
@@ -0,0 +1,103 @@
...
+  dash-tree-item[aria-expanded="false"] > label::before {
+    content: '+';
+  }
+  dash-tree-item[aria-expanded="true"] > label::before {
+    content: '-';
+  }
+  dash-tree-item[aria-expanded="true"] > dash-tree-group {
+    display: block;
+  }
+  dash-tree-item[aria-expanded="false"] > dash-tree-group {
```
robdodson: It may be worthwhile to add an expanded property to the treeitems, that would trigger the aria-expanded state when the element upgrades. This would be inline with the theory of keeping aria as an implementation detail that developers don't have to worry about

robdodson: fixed by adding expanded property

elements/dash-tree/demo.html
```diff
@@ -0,0 +1,103 @@
...
+    outline: 2px solid black;
+  }
+  .selected:not([aria-expanded]),
+  .selected:not([aria-expanded="true"]),
+  .selected[aria-expanded="true"] > label {
+    background: hotpink;
+  }
+</style>
+
+<!-- Possible PE: Upgrade from list/listem to tree/treeitem -->
```
robdodson: I should probably implement it

robdodson: fixed by adding the roles

elements/dash-tree/dash-tree.js
```diff
@@ -0,0 +1,457 @@
...
+      // A recursive function that visits every child and builds a list.
+      // This produces similar results to calling querySelectorAll,
+      // but allows for filtering of the children based on whether or not
+      // their parent is currently expanded.
+      function findTreeItems(node) {
+        for (let el of node.children) {
+          // If the child is a `DashTreeItem`, add it to the list of results.
+          if (isTreeItem(el)) treeItems.push(el);
+          // If it is not expanded, don’t descend.
+          // This should ignore any children and treat them as if they are
```
surma: s/should/will?

surma: Acutally, “This ignores all the children of the collapsed tree item...”

surma:
@alice: Would you mind taking a look at this element?

* [Page view](http://dash-elements.surma.link/7fac83eff3d3a76ed023ff19a166064333fb1914/dash-tree.html)
* [Isolated demo](http://dash-elements.surma.link/7fac83eff3d3a76ed023ff19a166064333fb1914/dash-tree_demo.html)

robdodson:
I added e2e and unit tests.

There are two spots where the tests were acting flakey so I put in a `.skip()`.
https://github.com/GoogleChrome/dash-elements/pull/29/files#diff-6d14add73c6f98296bd7d8b0a1769c60R233

https://github.com/GoogleChrome/dash-elements/pull/29/files#diff-6d14add73c6f98296bd7d8b0a1769c60R12

@surma your Selenium skills are better than mine, when you have a chance could you take a look and let me know if I'm doing anything obviously wrong there?

surma:
> could you take a look and let me know if I'm doing anything obviously wrong there?

On it

surma:
So it _seems_ (couldn’t verify by skimming Selenium’s code) that Selenium generates clicks at coordinates using the elements bounding box and clicking in the middle. So after your treeitem expanded, the click ended up on one of the children. I added code to click on the label instead.

Also it seems that on Safari `DOMElement.children` is a `NodeList` and can’t be used in with for...of, so I added `Array.from()`.

robdodson:
Just realized the way I'm doing labeling in this element is not reactive. _siiiiiiiigh_

robdodson:
ok @surma I _think_ this is ready for a final review. Phew, what a slog...

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  /**
+   * A helper to specifically identify `HowToTreeItemGroup` elements.
+   */
+  function isTreeItemGroup(node) {
+    return node.nodeName.toLowerCase() === 'howto-tree-item-group';
+  }
+
+  // `HowToTreeItemCounter` counts the number of treeItem instances
```
surma: Backticks around `treeItem`?

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,497 @@
...
+      // not currently supported by the `<label>` element, but hopefully
+      // they will be in the future. In the meantime, set the `aria-label`
+      // for the `HowToTreeItem`, equal to the `<label>` text.
+      // Without this labeling, the element's name will be computed based on
+      // its text content plus the text content of all of its children, making
+      // it so verbose as to be unusable.
+      if (!this.hasAttribute('aria-label')) {
+        let label = this.querySelector('label');
+        if (!label) {
+          console.error(`The first child of a <howto-tree-item> that ` +
```
surma: I have been throwing errors in these kinda situations. Logic being that the element cannot function if this requirement wasn’t met. What do we wanna settle for?

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,497 @@
...
+      // Before the elements starts booting, it waits for both
+      // the `<howto-tree-item>` and `<howto-tree-item-group>` to load.
+      Promise.all([
+        customElements.whenDefined('howto-tree-item'),
+        customElements.whenDefined('howto-tree-item-group'),
+      ]).then(_ => {
+        // Acquire all `HowToTreeItem`/`HowToTreeItemGroup` instances inside
+        // the element.
+        const treeItems = this._allTreeItems();
+        // If there are no treeItems, then the tree is empty. Abort.
```
surma: Backticks?

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,497 @@
...
+      // This produces similar results to calling querySelectorAll,
+      // but allows for filtering of the children based on whether or not
+      // their parent is currently expanded.
+      function findTreeItems(node) {
+        for (let el of Array.from(node.children)) {
+          // If the child is a `HowToTreeItem` or `HowToTreeItemGroup`, add it
+          // to the list of results.
+          if (isTreeItem(el))
+            treeItems.push(el);
+          // If it is a `HowToTreeIteMgroup` and it's collapsed, don’t descend.
```
surma: Capitalization

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,497 @@
...
+      this._toggleTreeItem(item);
+    }
+
+    /**
+     * Return the current active treeItem if there is one. Otherwise,
+     * return null.
+     */
+    _currentTreeItem() {
+      const currentTreeItem = this.querySelector('.active');
+      if (currentTreeItem)
```
surma: Sooo, just `return currentTreeItem`?

elements/howto-tree/howto-tree.js
robdodson: My thinking is that this kind of thing should actually be handled by a mutation observer. For instance, if someone changes the label's text or swaps in a new one, this code wouldn't run anymore.

I'd like to tackle the MO thing in a separate PR if that's ok

elements/howto-tree/howto-tree.js
robdodson: indeeeed

robdodson:
I'm getting some weird test failures after rebasing against master.

[This test](https://github.com/GoogleChrome/howto-components/blob/cdd7e49a34276531504c0298cdc2548d04cde1fa/elements/howto-tree/howto-tree.e2etest.js#L232) consistently fails on my local machine. It's an odd failure because if you look in the console it makes it seem like the methods aren't defined on the element. Checking the element's constructor shows HTMLElement instead of HowToTree. But then it's as if some of the handlers (like the click handler) have been installed but other parts of the script are missing. Super weird.

Also if you just move that test so it's before the previous test, everything works. :\

elements/howto-tree/howto-tree.js
surma: Sg

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,494 @@
...
+        this.removeAttribute('selected');
+    }
+
+    get selected() {
+      return this.hasAttribute('selected');
+    }
+
+    attributeChangedCallback(name, value) {
+      if (name === 'selected')
+        this.setAttribute('aria-selected', this.selected);
```
alice: Maybe a comment explaining why you're setting this here, analogous to the explanation about `aria-expanded`?

elements/howto-tree/howto-tree.js
```diff
@@ -0,0 +1,494 @@
...
+      // This handles the case where the element starts without an
+      // `[expanded]` attribute. In that scenario, the element would be set
+      // to `aria-expanded=false`.
+      this.setAttribute('aria-expanded', this.expanded);
+
+      // This first child should be a `<label>` element. Custom Elements are
+      // not currently supported by the `<label>` element, but hopefully
+      // they will be in the future. In the meantime, set the `aria-label`
+      // for the `HowToTreeItem`, equal to the `<label>` text.
+      // Without this labeling, the element's name will be computed based on
```
alice: Great explanation! :100:

elements/howto-tree/howto-tree.js
robdodson: fixed

surma:
I’ll take a look at the test failures.

surma:
Works for me locally and seems to work on Travis, too 😕

robdodson:
hm yeah it's weird. The tabs one always fails for me.

image

Seeeeems like it's maybe a race condition of some kind but I'm really not sure...

surma:
Can you `git clean -fxd` (that throws away everything!) and reinstall the node modules and see if it changes anything?

PR #30, opened 2017-02-28
Add howto-checkbox

robdodson:
For fun, here's a checkbox implementation. Just need to drop in comments and tests. I've got a lot of flights coming up, perfect time to write tests! ✈️

Something worth noting, I think this is the first of our elements to expose an attribute and a property. We may need to revisit the others at some point to give them similar APIs.

@surma @notwaldorf PTAL


robdodson:
ooh yeah i need to add labeling of some kind... will get to that... 😸

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    _onClick(event) {
+      this._toggleChecked();
+    }
+
+    _toggleChecked() {
+      const newState = !this.checked;
+      this.checked = newState;
+      this.dispatchEvent(new CustomEvent('check-changed', {
```
surma: Wouldn’t it be good to make this compatible to a normal `input[type=checkbox]` and fire a `change` event that bubbles?

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,77 @@
...
+      this.addEventListener('click', this._onClick);
+    }
+
+    _onKeyDown(event) {
+      // Don’t handle modifier shortcuts typically used by assistive technology.
+      if (event.altKey) return;
+
+      switch (event.keyCode) {
+        case KEYCODE.SPACE:
+          this._toggleChecked();
```
surma: Do you wanna do `[disabled]`, too?

robdodson:
@surma PTAL

robdodson:
Added the `disabled` property

surma:
You just... deleted my commits 😭 

I added the commit that adds component-specific assets to the output folder (in this case the SVGs).

robdodson:
@alice whenever you have a chance, if you could review the element (https://github.com/GoogleChrome/dash-elements/pull/30/files#diff-681557de3fabc0a0d6f8cb334aabc281) that would be awesooooome

elements/dash-checkbox/dash-checkbox.js
robdodson: should be fixed in current version

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,123 @@
...
+    constructor() {
+      super();
+    }
+
+    /**
+     * `connectedCallback` sets the initial role and checks to see if the
+     * user has predefined an `aria-checked` state or `tabindex`.
+     */
+    connectedCallback() {
+      this.setAttribute('role', 'checkbox');
```
robdodson: I guess technically I should check to see if the user has applied their own role and respect that

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,123 @@
...
+      this.dispatchEvent(new CustomEvent('checked-changed', {
+        detail: {
+          checked: this.checked,
+        },
+        bubbles: false,
+      }));
+    }
+
+    /**
+     * `attributeChangedCallback` watches for changes to the `checked`
```
robdodson: todo: add comment mentioning it watches `[disabled]` as well

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,123 @@
...
+   * The `DashCheckbox` exposes a single `checked` attribute/property for
+   * toggling its state. Changes to the `checked` property will also be
+   * reflected to an `aria-checked` attribute.
+   */
+  class DashCheckbox extends HTMLElement {
+    static get observedAttributes() {
+      return ['checked', 'disabled'];
+    }
+
+    constructor() {
```
azakus: No need to have `constructor` if you're just calling `super`.

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,123 @@
...
+     * user has predefined an `aria-checked` state or `tabindex`.
+     */
+    connectedCallback() {
+      this.setAttribute('role', 'checkbox');
+      if (!this.hasAttribute('aria-checked'))
+        this.checked = false;
+      if (!this.hasAttribute('tabindex'))
+        this.setAttribute('tabindex', 0);
+
+      this.addEventListener('keydown', this._onKeyDown);
```
azakus: The event listeners use `this`, but you didn't bind the function or use arrows. 👃

elements/dash-checkbox/images/checked-checkbox.svg
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="18px" height="18px" viewBox="0 0 18 18" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 41.2 (35397) - http://www.bohemiancoding.com/sketch -->
+    <title>checkbox</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Material-Design-Stickersheet" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="components" transform="translate(-396.000000, -226.000000)" fill="#009688">
```
azakus: Too much precision for an 18x18 image. Maybe use svgo?

elements/dash-checkbox/dash-checkbox.js
azakus: Guard this on the first `connectedCallback` call

elements/dash-checkbox/dash-checkbox.js
robdodson: do I need to bind it if the listener is always going to be called in the context of this class instance?

azakus: Nah, I just always do it, but that's not necessary 👍

elements/dash-tabs/dash-tabs.e2etest.js
```diff
@@ -102,6 +102,7 @@ describe('dash-tabs', function() {
     async function() {
       const lastTab =
         await this.driver.findElement(By.css('[role=tab]:last-of-type'));
+      // FIXME: This line is missing an await
```
robdodson: oh yeah @surma this was a note I left to remind  you that I think this tabs test needs to await reading the attributes.

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,121 @@
...
+     * and `disabled` attributes and reflects those to the underlying
+     * properties.
+     */
+    attributeChangedCallback(name, oldValue, newValue) {
+      this[name] = this.hasAttribute(name);
+    }
+
+    /**
+     * The `checked` property reflects its state to the `aria-checked`
+     * attribute.
```
alice: You're not doing full JSDoc with params and return types?

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,121 @@
...
+     */
+    set checked(isChecked) {
+      this.setAttribute('aria-checked', isChecked);
```
alice: Not supporting `aria-checked="mixed"` then? https://www.w3.org/TR/wai-aria-1.1/#aria-checked

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,121 @@
...
+    get checked() {
+      return this.getAttribute('aria-checked') === 'true';
+    }
+
+    /**
+     * The `disabled` property reflects its state to the `aria-disabled`
+     * attribute. A disabled checkbox will be visible, but no longer operable.
+     */
+    set disabled(isDisabled) {
+      this.setAttribute('aria-disabled', isDisabled);
```
alice: Does this have any bearing on focusability?

elements/dash-checkbox/dash-checkbox.js
robdodson: we haven't been. We have a custom parser that extracts the comments for side by side documentation.

elements/dash-checkbox/dash-checkbox.js
robdodson: nope

elements/dash-checkbox/dash-checkbox.js
robdodson: yeah I guess it should

elements/dash-checkbox/dash-checkbox.js
alice: Can I see what that looks like?

alice: nm found it. Ok, so no help on parameter types then.

elements/dash-checkbox/dash-checkbox.js
robdodson: fixed

elements/dash-checkbox/dash-checkbox.js
robdodson: Not currently. We could add something like that down the line. cc @surma

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -103,6 +103,11 @@
      */
     set disabled(isDisabled) {
       this.setAttribute('aria-disabled', isDisabled);
+      if (isDisabled) {
```
alice: Check or call this method in `connectedCallback()` as well?

robdodson: I'm a little sleepy but I _think_ we don't need to because it'll get triggered by `attributeChangedCallback` when the element boots if the user has added `[disabled]`. If the user has NOT added disabled, then what's in connectedCallback will happen first and set tabindex to 0

alice: So `attributeChangedCallback` gets called for all attributes on boot? Why do you handle any in `connectedCallback` then?

Maybe add a test for this anyway? i.e.
```html
<dash-checkbox aria-disabled="true">  <!-- assert this is not focusable -->
```

elements/dash-tabs/dash-tabs.e2etest.js
surma: Well... I guess the return value did not equal `true`, indeed 🤔 Thanks!

elements/dash-checkbox/dash-checkbox.js
surma: We could. Not sure it adds much value since we are not generating API documentation. 

@notwaldorf @robdodson Opinions?

elements/dash-checkbox/dash-checkbox.js
surma: +1 for test

robdodson: `attributeChangedCallback` only gets called for attributes listed in `observedAttributes`. It seems it only gets called if something is present so if I didn't put `disabled` on the element it would not be called.

robdodson:
I added two new methods to safely modify attributes without triggering attributeChangedCallback https://github.com/GoogleChrome/dash-elements/pull/30/files#diff-681557de3fabc0a0d6f8cb334aabc281R79

I'm wondering if folks would prefer that I always call these methods? For instance, `tabindex` is not an observedAttribute so it will never trigger attributeChangedCallback. But for consistency sake I could still call `_safelySetAttribute` when I modify it. What do you all think? @surma @alice

elements/dash-checkbox/dash-checkbox.js
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Reflecting a property to an attribute may potentially cause an infinite
+     * loop with `attributeChangedCallback`. These helpers set a flag to ensure
+     * that `attributeChangedCallback` returns instead of trying to set the
+     * underlying property again.
+     */
+    _safelySetAttribute(attr, value) {
+      if (this._safelyModifyingAttribute) return;
+      this._safelyModifyingAttribute = true;
+      this.setAttribute(attr, value);
```
surma: Is this guaranteed to work in all CE implementations? I.e. could `attributeChangedCallback` be called in a microtask?

Also, what about removing the attribute from `observedAttributes` before changing the value? Would that work?

surma: I realized you can’t change observedAttributes at runtime.

Looking at the [spec](https://w3c.github.io/webcomponents/spec/custom/#custom-element-reactions), it _seems_ like this implementation is fine. Not entirely sure tho. @azakus WDYT?

> The way in which custom element reactions are invoked is done with special care, to avoid running author code during the middle of delicate operations. Effectively, they are delayed until "just before returning to user script". This means that for most purposes they appear to execute synchronously, but in the case of complicated composite operations (like cloning, or range manipulation), they will instead be delayed until after all the relevant user agent processing steps have completed, and then run together as a batch.

robdodson:
k @surma @alice i think this is ready for review again

robdodson:
I think I've addressed all outstanding issues.

I added a label to the demo as a near term fix. Maybe a good next element to tackle would be a `<howto-label>`.

@surma PTAL. I'd love to merge this tomorrow if we can.

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,144 @@
...
+        },
+        bubbles: false,
+      }));
+    }
+
+    /**
+     * `attributeChangedCallback` watches for changes to the `checked`
+     * and `disabled` attributes and reflects their states to the corresponding
+     * ARIA attributes. It will be called at startup time if either attribute
+     * has been set. Because both `checked` and `disabled` are booleans, the
```
alice: Maybe pull "Because...present" into an inline comment?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,144 @@
...
+        default:
+          break;
+      }
+    }
+
+    /**
+     * The `checked` property reflects its state to the `checked`
+     * attribute.
+     */
+    set checked(isChecked) {
```
alice: Maybe move this (and all the other setters and getters) up earlier to make the `_toggleChecked()` -> `set checked()` -> `attributeChangedCallback()` sequence easier to follow?

elements/howto-checkbox/howto-checkbox.unittest.js
```diff
@@ -0,0 +1,31 @@
...
+
+    it('should add a `role` to the checkbox', function() {
+      expect(this.checkbox.getAttribute('role')).to.equal('checkbox');
+    });
+
+    it('should add a `tabindex` to the checkbox', function() {
+      expect(this.checkbox.getAttribute('tabindex')).to.equal('0');
+    });
+
+    it('should toggle `checked` and `aria-checked` when calling ' +
```
alice: Add a similar test for `checkbox.checked = false`/`true`?

elements/howto-checkbox/howto-checkbox.unittest.js
```diff
@@ -0,0 +1,31 @@
...
+      '`_toggleChecked`', function() {
+        expect(this.checkbox.checked).to.equal(false);
+        this.checkbox._toggleChecked();
```
alice: I assume this method isn't normally public..?

surma:
Closing this one in favor of #41 then :)

PR #38, opened 2017-04-04
s/panel/tabpanel

bkardell:
Panel isn't a role afaik, this wants to be tab panel

### DIFF
--- a/elements/dash-tabs/dash-tabs.js
+++ b/elements/dash-tabs/dash-tabs.js
@@ -311,7 +311,7 @@
     }
 
     connectedCallback() {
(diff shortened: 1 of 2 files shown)


surma:
According to [WAI-ARIA Best Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel) it is a role. Is that out-of-date?

bkardell:
where do you see panel?  you see that I am trying to change it *to* tabpanel, right?

surma:
> you see that I am trying to change it to tabpanel, right?
... No I didn’t. It’s 2am in Tokyo. Apologies! Thank you!

PR #41, opened 2017-04-13
Howto checkbox redux

robdodson:
This is a different approach on the checkbox. This version uses internal properties like `this._checked`. I feel this makes it easier to understand than the previous version which relied entirely on attributes.

I realize this is a departure from how we've been coding things but I really disliked all of the cleverness that the previous PR was relying on in order to do attribute/property reflection. I feel this version is just much more straightforward.


surma:
So you wanna land this _instead_ of #30? Or in addition to #30?

elements/howto-checkbox/demo.html
```diff
@@ -0,0 +1,38 @@
...
+    vertical-align: middle;
+    display: inline-block;
+    font-weight: bold;
+    font-family: sans-serif;
+    font-size: 20px;
+    margin-left: 8px;
+  }
+</style>
+
+<howto-checkbox aria-labelledby="#join-label"></howto-checkbox>
```
surma: `#` should be removed, right?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+      if (!this.hasAttribute('tabindex'))
+        this.setAttribute('tabindex', 0);
+
+      this._checked = false;
+      this._disabled = false;
+
+      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('click', this._onClick);
+    }
+
```
surma: `disconnectedCallback()`?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+        return;
+
+      switch (event.keyCode) {
+        case KEYCODE.SPACE:
+          this._toggleChecked();
+          break;
+        // Any other key press is ignored and passed back to the browser.
+        default:
+          return;
+      }
```
surma: `event.preventDefault()`?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    _onClick(event) {
+      this._toggleChecked();
+    }
+
+    /**
+     * `_toggleChecked` calls the checked setter and flips its state.
```
surma: Put backticks around `checked`

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+     */
+    _toggleChecked() {
+      if (this.disabled)
+        return;
+      this.checked = !this.checked;
+      this.dispatchEvent(new CustomEvent('change', {
+        detail: {
+          checked: this.checked,
+        },
+        bubbles: false,
```
surma: [Native `change` events bubble](https://jsbin.com/lotodok/1/edit?html,js,console,output), do we want to replicate that behavior or is there a good reason to not bubble?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * The `HowToCheckbox` exposes a single `checked` attribute/property for
+   * toggling its state. Changes to the `checked` property will also be
+   * reflected to an `aria-checked` attribute. Similarly, the `disabled`
+   * property is reflected to an `aria-disabled` attribute. This controls
+   * whether the element is operable or not.
+   */
+  class HowToCheckbox extends HTMLElement {
+    static get observedAttributes() {
+      return ['checked', 'disabled'];
```
surma: So here’s a fun thing: Native checkboxes _don’t_ react to changes of their `checked` attribute, but they _do_ react to a change of `disabled`. Is that insanity that we want to make more consistent or do we want to mimic the native element as closely as possible?

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+        return;
+      }
+      if (value === false && this.hasAttribute(attr)) {
+        if (isAriaAttribute)
+          this.setAttribute(attr, 'false');
+        else
+          this.removeAttribute(attr);
+        return;
+      }
+    }
```
surma: I think this can be simplified:

```js
const isAriaAttribute = attr.indexOf('aria-') === 0;
if (isAriaAttribute) {
  // aria-* attributes need strings "true" or "false" as values
  this.setAttribute(attr, value.toString());
  return;
}
if (value === false) this.removeAttribute(attr);
if (value === true && !this.hasAttribute(attr)) this.setAttribute(attr, '');
```

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * The `checked` property reflects its state to the `checked` and
+     * `aria-checked` attributes.
+     */
+    set checked(isChecked) {
+      if (this._checked === isChecked)
+        return;
+      this._checked = isChecked;
+      this._toggleAttribute('checked', isChecked);
```
surma: Again, this is unlike the native checkbox.

elements/howto-checkbox/howto-checkbox.js
surma: Also the current implementation is buggy in that it only takes the first branch if the attribute is _not_ present, which does not necessarily apply for aria-* attributes. In the current demo try: check, uncheck and check again. The second check won’t work because of this.

alice:
I think that's the plan, yeah

elements/howto-checkbox/howto-checkbox.js
robdodson: funny. Ummmmmmm I think let's just do the easy thing and be consistent.

elements/howto-checkbox/howto-checkbox.js
robdodson: My opinion on this comes from talking with Seb on the React team. He's of the opinion that elements should not bubble events, but instead their events should be handled by an immediate parent who then re-dispatches a more semantic event. I kinda like this notion so that's why I went with it here. I can add a comment explaining that and mention that it's a stylistic preference.

surma: Personally I think that rule of thumb is a little weird if that element is trying to replace a native element. But I can see where they are coming from. I’m fine with keeping it as is, but it warrants an explanation.

robdodson: yeah this is a strong opinion, weakly held. I could go either way.

surma: Your call, my preference goes for mimicking the native behavior.

robdodson:
K I think I'm done. For reals this time 😸 

The one thing I chose not to do is mimic the super bizarre behavior of `value` in the native checkbox. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value

This checkbox can't be submitted by a form anyway, so didn't seem necessary to mimic the weirdness that seems to have been caused by browsers extending `<input>`.

elements/howto-checkbox/howto-checkbox.js
robdodson: ah before I forget...

An important reason to reflect is for styling. A native checkbox has a `:checked` pseudo class, but there isn't a similar pseudo class for `disabled`. I wonder if that's why this behavior was chosen?

In our case, we don't have access to any pseudo classes so we need to reflect if folks want to style based on element state.

surma: Good point! Worth a comment imo

surma:
Just have that one remaining comment regarding not mimicking `value`. Feel free to do that later. Merge at will 🎉

surma:
LGTM

PR #42, opened 2017-04-19
Accordion – Attempt #2

surma:
High-level change list:

* The animation code is now much simpler and cleaner
* Using `<howto-accordion-heading>` for headings, `<section>` for content panels
* Using roving tab index
* Custom event from headings to start expansion logic

@robdodson PTAL :)


elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+     * `_allPanels` returns all the panels in the accordion. This function could
+     * memoize the result if the DOM queries ever become a performance issue.
+     * The downside of memoization is that dynamically added headings and panels
+     * will not be handled.
+     *
+     * This is a method and not a getter, because a getter implies that it is
+     * cheap to read while this method queries the DOM on every call.
+     */
+    _allPanels() {
+      return Array.from(this.querySelectorAll('howto-accordion-heading + *'));
```
robdodson: if the panels self apply `role=region` could you select that instead?

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+     */
+    _allHeadings() {
+      return Array.from(this.querySelectorAll('howto-accordion-heading'));
+    }
+
+    /**
+     * `_panelForHeading` returns the panel that the given heading controls.
+     */
+    _panelForHeading(heading) {
+      return heading.nextElementSibling;
```
robdodson: is it possible to assert that this next sibling is actually a region?

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+      // Put both the heading and the panel in the expanded state immediately.
+      // The animation function will hide the panel.
+      this._expandHeading(heading);
+      this._expandPanel(panel);
+
+      this._animateIn(panel);
+    }
+
+    _collapsePanel(panel) {
+      panel.setAttribute('aria-hidden', 'true');
```
robdodson: Seeing stuff like this makes me wonder if it's not better to just create a `<howto-accordion-panel>` element that could do all this stuff in response to a property or attribute being set. Right now similar logic lives inside the `<howto-accordion-heading>` so this element has a sort of split personality when it comes to its responsibilities.

elements/howto-accordion/demo.html
```diff
@@ -0,0 +1,44 @@
...
+  howto-accordion > *.animating {
+    transition: transform 0.3s ease-in-out;
+  }
+  howto-accordion-heading + .collapsed {
+    display: none;
+  }
+</style>
+
+<howto-accordion>
+  <howto-accordion-heading><button>Tab 1</button></howto-accordion-heading>
```
robdodson: If it's so important for the heading to have a button, yet you cant guarantee that someone using the element will put one in there, it actually could be a good reason to use shadow dom and have the element inject its own button.

I realize that opens the flood gate of shadow dom and all of its weirdness, but it's something to consider :D

robdodson:
Great job on this. A few things that got me thinking...

I wonder if it would be worth it to make the `<section>` it's own custom element. Then you wouldn't need to fiddle with its attributes from the parent accordion element. Right now you only set properties on the heading and it handles the rest. It seems like you'd want the same behavior for the section.

Also it got me wondering if this would be a good place to use Shadow DOM for a required element like the button. I realize Shadow DOM is its own can of worms but it feels like it's possibly a good fit here.

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+
+    /**
+     * `connectedCallback` gives unique IDs to all panels and headings, so it
+     * can then make headings and panels reference each other using
+     * `aria-controls` and `aria-labelledby`. Lastly, it considers the `expand`
+     *  attribute on the headers and adjusts the styling accordingly.
+     */
+    connectedCallback() {
+      // `<howto-accordion-headers>` emit a custom event when the heading was
+      // instructed to expand. The custom event is a nice abstraction expansion
```
surma: That’s because it’s not a sentence 😅

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    /**
+     * `_onHowtoChange` handles the custom `howto-change` event. The event’s
+     * target is the heading that has been instructed to expand (by click,
+     * keyboard input or by changing the `expanded` attribute or property).
+     */
+    _onHowtoChange(event) {
+      // If there’s an animation running, ignore the event.
+      if(this.classList.contains('animating')) return;
```
surma: You mean I should put the `return` on a separate indented line?:

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+      const heading = event.target;
+      const panel = this._panelForHeading(heading);
+      if(event.detail.isExpandedNow)
```
surma: I was convinced the google ESLint settings enforce something here, but apparently I was wrong. I’ll fix that in a separate PR as it’s gonna be big PR of just adding whitespace.

elements/howto-accordion/howto-accordion.js
surma: I have been flip-flopping back and forth on this. It’s really just an element. But if we can settled on the fact that something like ID generation and attribute management is enough to warrant a CE, I am more than happy to flop back :D

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    /**
+     * `connectedCallback()` sets up the role, event handler and initial state.
+     */
+    connectedCallback() {
+      if(!this.hasAttribute('role'))
+        this.setAttribute('role', 'heading');
+      const button = this.querySelector('button');
+      if(!button)
```
surma: I guess I am violating statelessness here. Will give this a think.

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+     * `attributeChangedCallback` processes changes to the `expanded` attribute.
+     */
+    attributeChangedCallback(name) {
+      // `expanded` is a boolean attribute it is either set or not set. The
+      // actual value is irrelevant.
+      const value = this.hasAttribute('expanded');
+
+      // Dispatch an event that signals a request to expand to the
+      // `<howto-accordion>` element.
+      this.dispatchEvent(
```
surma: How should I communicate with the parent, then? Direct function call? Seems to be going against CE best practices at a first glance.

I liked this because the event abstracts the actual cause. The event say “Something caused me to signal an intent to expand”. That could be a click, a keypress, an attributes change – doesn’t matter.

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+        new CustomEvent('howto-change', {
```
surma: I prefix’d it because it doesn’t really feel like the native `change` event. Maybe I’m wrong?

elements/howto-accordion/demo.html
surma: It would abstract a lot of weirdness. I am a little worried about the styling without JAvaScript, but I’ll give that a go as it will also address loads of your other comments.

elements/howto-accordion/howto-accordion.js
surma: This might be addresses by ShadowDOM

elements/howto-accordion/howto-accordion.js
robdodson: yeah

robdodson: and a space after if

elements/howto-accordion/howto-accordion.js
robdodson: ok :)

elements/howto-accordion/howto-accordion.js
robdodson: my personal opinion is to use a CE here. Especially if it means the CE could expose a property/attribute pair that, when set, triggers the corresponding aria stuff.

elements/howto-accordion/howto-accordion.js
robdodson: yeah if you used shadow dom you could either inject the button yourself or listen for it in a slotchange event. Since you always want a button, it seems like you'd just inject it yourself.

elements/howto-accordion/howto-accordion.js
robdodson: A click or keypress would dispatch an event, but an attribute change would not. The idea being that a click or keypress is a user driven action. Whereas an attribute change is something the program is doing.

For an attribute change I guess you could have the parent use a Mutation Observer. This might be a good use case to bring up with Domenic and Dominic over in the howto-components doc.

elements/howto-accordion/howto-accordion.js
robdodson: I don't think it's wrong necessarily. I just used `change` because I noticed I got freebie Angular directive support with ngModel when I did that. But this isn't a form element so it probably wouldn't use ngModel anyway.

`change` seems ok to me. At least if no better alternative presents itself.

elements/howto-accordion/howto-accordion.js
surma: Solved by having a CE for panels

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,533 @@
...
+        // Give all headings and panels a unique ID. Set up `aria-controls` and
+        // `aria-labelledby` attributes on headings and panels using those IDs.
+        headings.forEach(heading => {
+          // All buttons inside the `HowtoAccordionHeadings` are made
+          // unfocusable here. Only the first heading will be made focusable
+          // afterwards. This is necessary to implement roving tab index.
+          heading.querySelector('button').setAttribute('tabindex', -1);
+          const panel = this._panelForHeading(heading);
+
+          // Headings and panels need an ID so they can reference each other
```
surma: Solved by using ShadowDOM

elements/howto-accordion/howto-accordion.js
surma: Sold.

surma:
@robdodson @devnook PTAL :)

surma:
[Live preview](http://dash-elements.surma.link/a095e8cb6642453a35990b52ec470a4c3e960b2a/howto-accordion.html)

elements/howto-accordion/howto-accordion.e2etest.js
```diff
@@ -0,0 +1,95 @@
...
+
+  it('should expand a panel on click', async function() {
+    const lastHeading = await this.driver.findElement(By.css('[role=heading]:last-of-type'));
+    const lastPanelId = await lastHeading.getAttribute('aria-controls');
+    const lastPanel = await this.driver.findElement(By.id(lastPanelId));
+    expect(lastHeading.getAttribute('aria-expanded')).to.not.equal('true');
+
+    await lastHeading.click();
+    await helper.sleep(500);
+    expect(await lastPanel.getAttribute('aria-hidden')).to.equal('false');
```
robdodson: do you need to check that the expanded attributes changed as well?

elements/howto-accordion/README.md
```diff
@@ -0,0 +1,14 @@
+Accordions are a pattern to limit visible content by separating
+it into multiple panels. Any panel can be expanded or collapsed, giving
+the user control over which content is visible.
+
+By either clicking or by using the arrow keys the user changes the
+selection of the active heading. With enter or space the active headings
+can be toggled between expanded and collapsed state.
+
+The headings and the panels have the classes `expanded` or `collapsed`
```
robdodson: classes or attributes?

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+   * `<howto-accordion-panel>` and must be adjacent to a heading.
+   */
+  class HowtoAccordion extends HTMLElement {
+    constructor() {
+      super();
+    }
+
+    /**
+     * `connectedCallback` hooks up the even listeners and considers the
+     * `expand`  attribute on the headers to adjust their styling accordingly.
```
robdodson: `expanded`. 

Also you've got two spaces after `expanded`

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+     */
+    connectedCallback() {
+      // `<howto-accordion-headers>` emit a custom event when the heading is
+      // instructed to expand. The custom event is a nice abstraction because
+      // expansion can be triggered by clicks, keyboard input and attribute
+      // and property changes alike.
```
robdodson: I don't think attribute/property changes should dispatch events.

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+      this.addEventListener('change', this._onChange);
+      // The element also implements roving tab index to switch focus between
```
robdodson: Link for roving tabindex (and make tabindex one word)

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+      // the headers. Therefore key presses are intercepted.
+      this.addEventListener('keydown', this._onKeyDown);
+
+      // TODO: Set up MutationObserver to listen for `expand` attribute on the
```
robdodson: Cool leaving the TODO in for now?

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+  // button’s accessible functionality is preserved while still allowing
+  // the developer to freely style the headings.
+  //
+  // Another advantage is focus management. If the button inside ShadowDOM has
+  // focus, `document.activeElement` returns the containing
+  // `<howto-accordion-heading>` element rather than the button itself.
+  const shadowDOMTemplate = document.createElement('template');
+  shadowDOMTemplate.innerHTML = `
+    <style>
+      :host > button {
```
robdodson: do you need :host > button? I think `button` might work

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+  /**
+   * `HowtoAccordionHeading` is the element for the headings in the accordion.
+   * Accordion to the WAI ARIA Best Practices, each heading needs to wrap a
+   * `<button>`. This element puts that element in the ShadowDOM, as it is more
+   * convenient to use and doesn’t make server-side rendering or styling more
+   *  problematic. This element dispatches a `howto-accordion-change` event when
+   * it is supposed to expand.
+   *
+   * Clicking the button or pressing space or enter while the button has focus
+   * will expand the heading. Changing the `expand` attribute or property will
```
robdodson: `expanded`

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+        // Acquire all headings inside the element that need to be set up.
+        const headings = this._allHeadings();
+
+        // Give all headings and panels a unique ID. Set up `aria-controls` and
+        // `aria-labelledby` attributes on headings and panels using those IDs.
+        headings.forEach(heading => {
+          // All buttons inside the `HowtoAccordionHeadings` are made
+          // unfocusable here. Only the first heading will be made focusable
+          // afterwards. This is necessary to implement roving tab index.
+          heading._shadowButton.setAttribute('tabindex', -1);
```
robdodson: Don't think we should be exposing our shadow dom internals to the outside world.

If you need to make all the children of a shadow root unfocusable you can put tabindex=-1 on the hosting element.

https://jsbin.com/dadihup/1/edit?html,output

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+          const panel = this._panelForHeading(heading);
+
+          // Make headings and panels reference each other
+          // with the `aria-labelledby` and `aria-controls` attributes.
+          heading.setAttribute('aria-controls', panel.id);
+          panel.setAttribute('aria-labelledby', heading.id);
+
+          // Assign the appropriate roles to panels. Headings are custom
+          // elements and set their role in their own `connectedCallback`.
+          if (!panel.hasAttribute('role'))
```
robdodson: This can be removed because panels are CEs as well now

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+            panel.setAttribute('role', 'region');
+        });
+        // Make the first heading focusable.
+        headings[0]._shadowButton.setAttribute('tabindex', 0);
```
robdodson: I'd suggest removing the `tabindex=-1` from the first heading

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+          return;
+      }
+
+      // The browser might have some native functionality bound to the arrow
+      // keys, home or end. The element calls `preventDefault` to prevent the
+      // browser from taking any actions.
+      event.preventDefault();
+      // Make the currently focused heading unfocusable, then make the new
+      // heading focusable and give focus to it.
+      currentHeading._shadowButton.setAttribute('tabindex', -1);
```
robdodson: Add/remove `tabindex=-1` from the headings themselves rather than interacting with their shadow dom internals

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,591 @@
...
+      newHeading._shadowButton.setAttribute('tabindex', 0);
+      newHeading._shadowButton.focus();
```
robdodson: To be able to focus the element I think you'll need to use the incredibly esoteric `delegatesFocus` flag. It's explained in the 2nd half of this handling focus section.
https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom#focus

Here's an example. Note that I have to remove the focus ring from the div that's my shadow host.
https://jsbin.com/xiyate/2/edit?html,output

elements/howto-accordion/howto-accordion.js
robdodson: Maybe use css containment?
https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom#containment

elements/howto-accordion/howto-accordion.js
robdodson: oh actually nevermind. If you use the `delegatesFocus` flag, you can have tabindex=0 on the heading, and it'll just pass through to the first focusable child.

elements/howto-accordion/howto-accordion.js
surma: Eh, that was a left-over comment :-/

elements/howto-accordion/howto-accordion.js
surma: Yes. That’s one of the major cross-component cleanup items I think

elements/howto-accordion/howto-accordion.js
surma: Good call!

elements/howto-accordion/howto-accordion.e2etest.js
surma: YES!

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,601 @@
...
+
+      // The browser might have some native functionality bound to the arrow
+      // keys, home or end. The element calls `preventDefault` to prevent the
+      // browser from taking any actions.
+      event.preventDefault();
+      // Make the currently focused heading unfocusable, then make the new
+      // heading focusable and give focus to it.
+      currentHeading.setAttribute('tabindex', -1);
+      newHeading.setAttribute('tabindex', 0);
+      newHeading._shadowButton.focus();
```
robdodson: focus the heading not the shadowbutton

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,601 @@
...
+
+    /**
+     * `_expandHeading` puts the given heading in the expanded state, without
+     * any animation.
+     */
+    _expandHeading(heading) {
+      heading.expanded = true;
+    }
+
+    _collapsePanel(panel) {
```
robdodson: For some reason I really want this to be next to _collapsePanel.

elements/howto-accordion/howto-accordion.js
```diff
@@ -0,0 +1,601 @@
...
+    }
+
+    /**
+     * `attributeChangedCallback` processes changes to the `expanded` attribute.
+     */
+    attributeChangedCallback(name) {
+      // `expanded` is a boolean attribute it is either set or not set. The
+      // actual value is irrelevant.
+      const value = this.hasAttribute('expanded');
+      this.setAttribute('aria-expanded', value);
```
robdodson: move to shadow button

elements/howto-accordion/howto-accordion.js
surma: I had them grouped by `collapseXXX` and `expandXXX`, but guess those times are over :D Will fix

surma:
@robdodson final PTAL? :D

robdodson:
Remove aria-hidden stuff and i think we're good

surma:
@robdodson Is this the moment?

robdodson:
bacon and eggs dancing

PR #46, opened 2017-04-24
Checkbox lazy props

robdodson:
This is a branch of #45 so you can ignore most of the changes. The interesting bits are in [this commit](https://github.com/GoogleChrome/howto-components/commit/931effd5da4d4da80f47bedf80308de36c606045).

This is based on the pattern recommended by Scott Miles from the Polymer team. The idea is if an element's definition is lazy loaded, it can still recover if someone has set a property value on the unupgraded instance.

I'd like to write a test for this but I'm not quite sure how to do that with the current test setup. [Here's a jsbin that illustrates what I'd like to do](https://jsbin.com/malavet/edit?html,console,output).

### DIFF
--- a/elements/howto-checkbox/howto-checkbox.e2etest.js
+++ b/elements/howto-checkbox/howto-checkbox.e2etest.js
@@ -77,7 +77,7 @@ describe('howto-checkbox pre-upgrade', function() {
     return this.driver.get(`${this.address}/howto-checkbox_demo.html?nojs`);
   });
 
-  it('should upgrade attributes on upgrade',
+  it('should handle attributes set before upgrade',
     async function() {
       await this.driver.executeScript(_ =>
         window.expectedCheckbox = document.querySelector('howto-checkbox')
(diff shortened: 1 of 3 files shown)


surma:
ooh! Lovely pattern! I added an example test in ebe14e6d175f6f23fb605780551d6236f99bbd9a. WDYT?

We’ll have to make a pass over the existing components and add that behavior if needed to be consistent.

robdodson:
Cool thank you for doing that! I'm also going to add one which sets the property. When the element boots up I believe `attributeChangedCallback` will always be called so setting the attribute should get detected automatically. Setting the property wouldn't though, and that's the real tricky bit that this pattern hopefully addresses. The use case is a framework like Angular binding to the property before the element is ready.

robdodson:
ok added a properties test as well. Thanks again for hooking this up!

surma:
So this LGTM. Merge at will after rebase from my side :)

PR #47, opened 2017-04-26
Adding initial version of howto-menu

devnook:


elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
...
+  howto-menuitem {
+    display: inline-block;
+    width: 100%;
+    padding: .2em .5em;
+    border: 1px solid lightgrey;
+  }
+
+  /* The menu is closed. */
+  howto-menu[aria-hidden="true"] {
+    visibility: hidden;
```
surma: Any reason why you are choosing `visibility: hidden` over `display: none`?

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
+<!doctype html>
+
+<style>
+  /* The tooltip is by default unstyled. */
+  howto-menu {
+    display: block;
```
surma: With the implicit `position: static` the menu will push down following content when opened and re-layout the entire page. I think `position: absolute` & some JavaScript to set the position might be worth a try.

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /* The menu is open. */
+  howto-menu[aria-hidden="false"] {
+    visibility: visible;
+  }
+</style>
+
+<div class="text">
+   <button aria-label="menu" id="menu-btn">open menu</button>
```
surma: I don’t think a `<button>` needs a label.

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
...
+   <howto-menu aria-labelledby="menu-btn" aria-hidden="true">
```
surma: For progressive enhancement it is _probably_ better to have the element set the `aria-hidden` on boot rather than putting it in the markup. When JavaScript is not enabled, the menu will be hidden from screenreaders.

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
...
+     <howto-menuitem>second option</howto-menuitem>
+     <howto-menuitem>other option</howto-menuitem>
+     <howto-menuitem>yet another option</howto-menuitem>
+   </howto-menu>
+</div>
+
+<script>
+  document.getElementById('menu-btn').addEventListener('click', e => {
+    var menu = document.querySelector('howto-menu');
+    menu.toggle();
```
surma: So far we have been implementing over APIs with attributes as it allows to integrate with frameworks like Angular well. So instead you’d do
```js
if (menu.hasAttribute('expanded')
  menu.removeAttribute('expanded');
else
  menu.setAttribute('expanded', '');
```

or something along those lines.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+ * A menu is a widget that offers a list of choices to the user,
+ * such as a set of actions or functions. A menu is usually opened,
+ * or made visible, by activating a menu button, choosing an item in a menu
+ * that opens a sub menu, or by invoking a command, such as Shift + F10 in
+ * Windows, that opens a context specific menu.
+ *
+ * The element that opens the menu is referenced with aria-labelledby.
+ * TODO(devnook): Should it be aria-labelledby or aria-describedby?
+ *
+ * See: https://www.w3.org/TR/wai-aria-practices-1.1/#menu
```
surma: That’s a pretty good idea. We should always put that link there!

surma: Opened #48

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+
+ class HowtoMenu extends HTMLElement {
+
+   /**
+    * The constructor does work that needs to be executed _exactly_ once.
+    */
+   constructor() {
+     super();
+
+     this.KEY_CODES_ = {
```
surma: So far we’ve been putting our code in an IIFE and have the key codes outside the class ([example](https://github.com/GoogleChrome/howto-components/blob/f9f855beeaade5d3d01c0e9f91b861bbf092fe5b/elements/howto-radio-group/howto-radio-group.js#L28))

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+       ARROW_DOWN: 40,
+       ARROW_UP: 38,
+       ESCAPE: 27,
+     };
+     this.children_ = null;
```
surma: Private fields have their underscore at the start.

surma: We are trying to be stateless, so we have been avoiding variables like this that “cache” children. Currently the menu would not know about dynamically added menu points.

It’s okay to query for all the children whenever you need ([example](https://github.com/GoogleChrome/howto-components/blob/f9f855beeaade5d3d01c0e9f91b861bbf092fe5b/elements/howto-radio-group/howto-radio-group.js#L156)). Alternatively, you could set up a MutationObserver, although I don’t see the need for that complexity.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+     // TODO: Consider better naming?
+     this.parent_ = document.getElementById(
```
surma: Same.

Regarding naming: “Referencee”? “Opener”?

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+   }
+
+   moveFocus_(fromEl, toEl) {
+     fromEl.setAttribute('tabindex', -1);
+     toEl.setAttribute('tabindex', 0);
+     toEl.focus();
+   }
+
+   handleBlur_(e) {
+     e.target.setAttribute('tabindex', -1);
```
surma: I am not sure if `blur` bubbles, but if someone does crazy stuff like putting an `<input>` in their menu, this might be do unexpected things. I think a check if `e.target` is actually a `<howto-menuitem>` should suffice.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+           }
+         }
+         break;
+       case e.keyCode === this.KEY_CODES_.ESCAPE:
+         // If escape, exit the menu.
+         el.setAttribute('tabindex', -1);
+         this.parent_.focus();
+         this.toggle();
+         break;
+       // TODO: What should Tab key do?
```
surma: Should be left to the browser. All keypresses handled by you should call `e.preventDefault()`.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+         this.getAttribute('aria-labelledby'));
+   }
+
+   moveFocus_(fromEl, toEl) {
+     fromEl.setAttribute('tabindex', -1);
+     toEl.setAttribute('tabindex', 0);
+     toEl.focus();
+   }
+
+   handleBlur_(e) {
```
surma: Call it `event`.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+     e.target.setAttribute('tabindex', -1);
+   }
+
+   handleKeyDown_(e) {
+     let el = e.target;
+     switch (true) {
```
surma: I think I prefer a `switch` on `e.keyCode` and handling the letter key range in a separate `if` before the switch.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+       case e.keyCode === this.KEY_CODES_.ARROW_DOWN:
+         // If arrow down, move to next item. Wrap if necessary.
+         let next = el.nextElementSibling || this.firstElementChild;
```
surma: Move this into a getter á la [this](https://github.com/GoogleChrome/howto-components/blob/f9f855beeaade5d3d01c0e9f91b861bbf092fe5b/elements/howto-radio-group/howto-radio-group.js#L197)

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+     this.setAttribute('aria-hidden', !makeOpen);
+     if (makeOpen) {
+       // Set focus to the first item.
+       this.firstElementChild.setAttribute('tabindex', 0);
+       this.firstElementChild.focus();
+     }
+   }
+ }
+
+ window.customElements.define('howto-menu', HowtoMenu);
```
surma: Where is `howto-menuitem`?

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,43 @@
...
+  /* The menu is open. */
+  howto-menu[aria-hidden="false"] {
+    visibility: visible;
+  }
+</style>
+
+<div class="text">
+   <button aria-label="menu" id="menu-btn">open menu</button>
+   <howto-menu aria-labelledby="menu-btn" aria-hidden="true">
+     <howto-menuitem>first option</howto-menuitem>
```
surma: This CE should be called `<howto-menu-item>`

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+     });
+     this.removeEventListener('keydown', this.handleKeyDown_);
+   }
+
+   /**
+    * Open the menu if it was closed and vice versa.
+    */
+   toggle() {
+     // If the menu is hidden, show it.
+     let makeOpen = this.getAttribute('aria-hidden') === 'true';
```
surma: Probably nice to have a (private) method with a meaningful name for this.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,115 @@
...
+     this.setAttribute('aria-hidden', !makeOpen);
+     if (makeOpen) {
+       // Set focus to the first item.
+       this.firstElementChild.setAttribute('tabindex', 0);
```
surma: I’d prefer an `_open()` and a `_close()` method that you call into.

elements/howto-menu/demo.html
devnook: I wanted to keep the elements as minimal as possible, therefore was not really styling them (I focused on accessibility first). But you're right, if we want to make the elements performant too they should have some visuals too:) I'd say let's do it in next PR, together with animating etc.

elements/howto-menu/demo.html
devnook: My thinking was that the developer might want to choose whether a menu is visible on the page initially or hidden away. This said, now that I read again the wai-aria guidelines, a visible menu would be rather called a "menubar"... So it would be a different element I guess.

elements/howto-menu/howto-menu.js
devnook: Opener yay!

elements/howto-menu/howto-menu.js
devnook: Removed

devnook:
PTAL. 

Most of the issues are addressed:) Let's add more styling/animations in a next PR. 
Regarding branches, even as a collaborator, I don't think I have write access to create branches in the repo (I get 403).

surma:
That’s because I lied and you didn’t have access >.> I forgot to check, sorry. Fixed now :)

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+
+  class HowtoMenu extends HTMLElement {
+
+    /**
+     * The constructor does work that needs to be executed _exactly_ once.
+     */
+    constructor() {
+      super();
+      this._handleBlur = this._handleBlur.bind(this);
+      this._opener = document.getElementById(
```
surma: You only use `_opener` in `_close`, so I think you should just query for it there instead of caching it here.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+         this.getAttribute('aria-labelledby'));
+    }
+
+    static get observedAttributes() {
+      return ['aria-hidden'];
```
surma: Rob and I have been trying to use ARA just as an implementation detail. Make your own attributes (`expanded`?) and set `aria-hidden` accordingly instead.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+
+    /**
+     * A getter for the first child which is a menuitem.
+     */
+    get firstMenuItem() {
```
surma: Should probably be private

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+      return this.querySelector('[role^="menuitem"]:first-of-type');
+    }
+
+    /**
+     * A getter for the last child which is a menuitem.
+     */
+    get lastMenuItem() {
```
surma: Same

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+      return this.querySelector('[role^="menuitem"]:last-of-type');
+    }
+
+    /**
+     * Moves the browser focus from one element to another.
+     * @param {Element} fromEl Element to have focus removed.
```
surma: Hm, we haven’t used proper JSDoc so far. Since my parser currently doesn’t understand it, I am not sure it adds value. The names are pretty descriptive, too. @robdodson WDYT?

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * A getter for the last child which is a menuitem.
+     */
+    get lastMenuItem() {
+      return this.querySelector('[role^="menuitem"]:last-of-type');
+    }
+
+    /**
+     * Moves the browser focus from one element to another.
```
surma: Probably worth mentioning “roving tabindex” here.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+
+       default:
+         break;
+      };
+      // If letter key, move to an item which starts with that letter.
+      if (event.keyCode > 64 && event.keyCode < 91) {
+        for (let i = 0, child; child = this.children[i]; i++) {
+          if (child.innerText.trim()[0] === event.key) {
+            event.preventDefault();
+            this._moveFocus(el, child);
```
surma: and `break`, no?

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,39 @@
...
+
+  /* The menu is closed. */
+  howto-menu[aria-hidden="true"] {
+    display: none;
+  }
+</style>
+
+<div class="text">
+   <button id="menu-btn">open menu</button>
+   <howto-menu aria-labelledby="menu-btn">
```
robdodson: The goal with the label is to help the user understand the purpose of the menu. If you want to continue using a button I'd suggest changing the text to `"options"` instead of `"open menu"`.

elements/howto-menu/howto-menu.js
robdodson: I'd suggest creating a boolean property/attribute pair called `.opened` and `[opened]`.

Using `expanded` makes me think it will also set `aria-expanded` but I don't think we need to do that here. Instead we could reserve `expanded`/`aria-expanded` just for submenu items.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
+/*
+ * A menu is a widget that offers a list of choices to the user,
+ * such as a set of actions or functions. A menu is usually opened,
+ * or made visible, by activating a menu button, choosing an item in a menu
+ * that opens a sub menu, or by invoking a command, such as Shift + F10 in
+ * Windows, that opens a context specific menu.
+ *
+ * The element that opens the menu is referenced with aria-labelledby.
+ * TODO(devnook): Should it be aria-labelledby or aria-describedby?
```
robdodson: In the case of a menu button it should be `aria-labelledby`.

robdodson: You could additionally add an `aria-describedby` field to help explain the purpose of the menu.

example:

label: "Options"
describedby: "Configuration for your personal twitter account"

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+  };
+
+  class HowtoMenu extends HTMLElement {
+
+    /**
+     * The constructor does work that needs to be executed _exactly_ once.
+     */
+    constructor() {
+      super();
+      this._handleBlur = this._handleBlur.bind(this);
```
robdodson: We use the convention `_onBlur`

elements/howto-menu/howto-menu.js
robdodson: yeah just needs an underscore, `_firstMenuItem`

elements/howto-menu/howto-menu.js
robdodson: For reference: https://github.com/GoogleChrome/howto-components/blob/master/elements/howto-radio-group/howto-radio-group.js#L8

elements/howto-menu/howto-menu.js
robdodson: yeah i think you can remove this

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+     * Sets up keyboard interactions for menu and its items.
+     */
+    connectedCallback() {
+      // Make children unfocusable by default.
+      const children = Array.from(this.children);
+      children.forEach(el => {
+        el.setAttribute('tabindex', -1);
+        el.addEventListener('blur', this._handleBlur);
+      });
+      this.setAttribute('role', 'menu');
```
robdodson: Check if role is already set.

```
if (!this.hasAttribute('role'))
  this.setAttribute('role', 'menu');
```

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+        }
+        prev = prev.previousElementSibling;
+      }
+      return null;
+    }
+
+    /**
+     * Opens the menu.
+     */
+    _open() {
```
robdodson: I'd move this, and the work done in `_close` into the `attributeChangedCallback` after you create the `opened` attribute/property.

Take a look at how checkbox does it: https://github.com/GoogleChrome/howto-components/blob/master/elements/howto-checkbox/howto-checkbox.js#L122

That's the pattern we're trying to stick to. Property setters and getters just reflect to attributes and any side effects take place in the `attributeChangedCallback`. This is what Chrome eng suggested we do.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,216 @@
...
+      }
+    }
+
+    /**
+     * Sets up keyboard interactions for menu and its items.
+     */
+    connectedCallback() {
+      // Make children unfocusable by default.
+      const children = Array.from(this.children);
+      children.forEach(el => {
```
robdodson: I think having a `reset()` method similar to dash-tabs might be good here.
https://github.com/GoogleChrome/howto-components/blob/master/elements/howto-tabs/howto-tabs.js#L202

That way we could be consistent across elements.

elements/howto-menu/howto-menu.e2etest.js
```diff
@@ -0,0 +1,123 @@
...
+        return document.activeElement === document.querySelector('#menu-btn');
+      `);
+      expect(focusedMenuBtn).to.equal(true);
+      const closedMenu = await this.driver.executeScript(`
+        const menu = document.querySelector('howto-menu');
+        return menu.getAttribute('aria-hidden') === 'true';
+      `);
+      expect(closedMenu).to.equal(true);
+    }
+  );
```
robdodson: Also need to make sure the menu closes when the user tabs out of it.

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,39 @@
...
+     <span role="menuitem">first option</span>
+     <span role="menuitem">second option</span>
+     <span role="menuitem">other option</span>
```
robdodson: I'd really like to see an example of at least one nested menu item that uses `aria-haspopup`.

In general we've tried to avoid developers needing to write out ARIA attributes themselves, so it might be a good use case to create a `<howto-menu-item>` element that automatically self applies `aria-haspopup` if it has children.

elements/howto-menu/howto-menu.js
devnook: Sorry, old habits from my Closure days;)

elements/howto-menu/howto-menu.js
devnook: How about using built in 'hidden' attribute instead? Do you think it would be too much overload on the term?

elements/howto-menu/demo.html
devnook: I plan to add menuitem, menuitemradio and menuitemcheckbox. Also, there should be a demo where all of them work together. Is it ok to do it in a separate PR though?

surma: sg

elements/howto-menu/howto-menu.js
devnook: I'm not sure I got it correctly - you want the body of the _open and _close to be moved back  to  attributeChangedCallback? They are already called from there, just with the 'hidden' property and not 'opened'.

devnook:
PTAL

Most issues are addressed, apart from attributeChangedCallback (need a clarification from @robdodson there).

Also, tests seem broken, but it happens also on master - do you know where's the issue?

surma:
Master works for me (and Travis) just fine. Since this is in your repo, I can’t fix it for you, but a rebase to master should do it :)

devnook:
I'm abandoning this PR in favour of #78 (Somehow I'm unable to fix tests on this branch).

PR #49, opened 2017-04-26
Adding howto-menu-button element

devnook:


surma:
Is this really a component of its own? I think it make more sense rolling it `<howto-menu>` (#47) and using/testing them together.

devnook:
Actually this is something I wanted to talk about. I started developing them together, but then realized a few things:

1. Menu can be used without a menu button. E.g. it can be opened from a menubar, or by activating an option in another menu (a submenu), or by hitting a hotkey (context menu).
2. Putting many elements in a one demo might make it less clear which one does what. Here we'd end up with menu, menu button, menubar, menuitem, menuitemradio, menuitemcheckbox. That's quite a few. But maybe it's other way round and it's good to have them all interact together? 
3. (Less important) Current setup maps 1 js file to one demo, so they would need to live in 1 file (or we change the doc build process).

I'm still on the fence about it. Maybe a possible approach would be to have each separate js file and demo, but then have also 1 demo that pulls it all together... 

Share your thoughts:)

surma:
My thoughts: I think having them separate is confusing. Here you have one CE and try to turn some vanilla markup into a menu, and the other PR is the other way around. I’d rather have one fully polished setup which will also make the code much easier to read. That is my biggest argument btw. The code would be simpler and easier to understand.

I see your point about menus being usable without a button, but that’s kinda the beauty of the project – we don’t need to be universal or fully generic, as long as we’re covering the concepts. 

> Putting many elements in a one demo might make it less clear which one does what. Here we'd end up with menu, menu button, menubar, menuitem, menuitemradio, menuitemcheckbox. 

The subtypes are a good counter point. How about this: You do `<howto-menu> and `<howto-menubutton>` and the rest remain vanilla elements that are automatically decorated appropriately by `<howto-menu>`. Does that sound feasible?

I could add support for multiple files, but not sure it’s necessary. We have multiple CEs in `<howto-tabs>` or `<howto-radio-group>` already.

@robdodson WDYT?

surma:
Closing this in favor of #79

PR #81, opened 2017-05-04
Update README.md

robdodson:
"“HowTo: Components” **(sometimes stylized as components)** is a subsection on Web Fundamentals Architecture "

This statement was kind of confusing for me. I think the README is a bit more clear if we just cut it.


robdodson:
To clarify, I saw that it meant to say `<HowTo>` components, but I still think that's confusing :D

Issue #57, opened 2017-04-27
Elements should not rely on ARIA attributes for styling or behavior.


robdodson:
In earlier elements like `<howto-radio-group>`, I relied on `aria-*` for styling and behavior. Instead, ARIA should be an implementation detail.

So instead of styling `aria-checked` I would expose a `checked` attribute/property on the element, and style that. Setting `checked` would, as a side effect, also set `aria-checked`. We should use this pattern throughout.

Issue #55, opened 2017-04-27
All elements should use the lazy properties pattern.


robdodson:
Demonstrated in [this PR](https://github.com/GoogleChrome/howto-components/pull/46).

Make sure any property set on the instance before the element has upgraded will get picked up.

PR #86, opened 2017-05-08
Add MO & Lazy Props to Accordion

surma:
@robdodson @devnook PTAL! Esp. at the architecture around MO!

When merged, set a checkmark in #55, #57, #75 & #85


robdodson:
I _think_ it looks ok but I'm not a mutation observer expert. @azakus I'm sure your'e super busy but if you have any spare cycles to review this mutation observer implementation for an accordion that would be 🎂

PR #87, opened 2017-05-09
Add lazy proprs to Tabs

surma:
Interestingly, no need for MO as I can fully rely on the implicit ShadowDOM MO and the `slotchange` event.

When merged, set a checkmark in #55 

@devnook @robdodson PTAL :)


robdodson:
I'd suggest adding a `.trim()` to your template innerHTML call.

robdodson:
LGTM

PR #90, opened 2017-05-09
Streamline Tooltip

surma:
PTAL!

When merged, set a checkmark in #55, #57 and #85

### DIFF
--- a/elements/howto-tooltip/demo.html
+++ b/elements/howto-tooltip/demo.html
@@ -20,28 +20,8 @@ limitations under the License.
     background-color: #616161;
     color: #fff;
     border-radius: 3px;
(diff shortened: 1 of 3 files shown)


surma:
PS: Yes, I removed the animation for simplicities sake for now. I’ll do another PR for animations once we are in a better state universally.

robdodson:
small thing, can you wrap the `return` on line 46? Otherwise LGTM. Merge at your leisure ⛵️

surma:
>  can you wrap the return on line 46? 

Yes. I’ll also see if I can add that to eslint

surma:
Found the rule, will make separate commit!

PR #94, opened 2017-05-24
First pass at checkbox guide

robdodson:
@surma PTAL

PR #95, opened 2017-05-25
Proper intro section for tooltip

surma:
Like #94 but for tooltip :)

### DIFF
--- a/elements/howto-tooltip/README.md
+++ b/elements/howto-tooltip/README.md
@@ -1,7 +1,37 @@
+## Summary
+
 A tooltip is a popup that displays information related to an element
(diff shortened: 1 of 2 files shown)


robdodson:
This LGTM. Though I'm now definitely thinking that we should extract some of the common guide bits into a separate doc because it's going to be difficult if we want to update the language in one, to have to go to each guide and update it there as well.

surma:
Agreed. Let’s do that when you get back. We can launch these two elements separately for now :)

I’ll make adjustments according to the changes we made to your PR

robdodson:
sgtm!

surma:
WOOOOO 🎉

PR #98, opened 2017-07-27
One more attempt at the menu.

devnook:


elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+    ESCAPE: 27,
+    TAB: 9,
+    UP: 38,
+  };
+
+  class HowtoMenu extends HTMLElement {
+    /**
+     * A getter for the first child which is a menuitem.
+     */
+    get _firstMenuItem() {
```
surma: I think getters should only be used if they run in O(1). But I am happy to put that up for discussion.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+      }
+      return null;
+    }
+
+    /**
+     * Checks if a node is a 'menuitem', 'menuitemcheckbox' or 'menuitemradio'.
+     */
+    _isMenuItem(node) {
+      let ariaRoles = ['menuitem', 'menuitemcheckbox', 'menuitemradio'];
+      return ariaRoles.indexOf(node.getAttribute('role') > -1);
```
surma: `ariaRoles.includes(node.getAttribute('role'))`

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Handles keyboard interaction.
+     */
+    _onKeyDown(event) {
+      let el = event.target;
+      switch (event.keyCode) {
+        case KEYCODE.DOWN:
+          // If arrow down, move to next item. Wrap if necessary.
+          event.preventDefault();
+          let next = this._nextMenuItem(el) || this._firstMenuItem;
```
surma: ... this is actually super elegant. We should use this pattern in some of the other elements. cc @robdodson

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+       case KEYCODE.TAB:
+         // If escape, close the menu.
+         event.preventDefault();
+         this.opened = false;
+         break;
+       default:
+         break;
+      }
+      // If letter key, move to an item which starts with that letter.
+      if (event.keyCode > 64 && event.keyCode < 91) {
```
surma: Add constants for `64` and `91` to `KEYCODE`

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+        for (let i = 0, child; child = this.children[i]; i++) {
```
surma: I have an itch to use functional programming here, but I’ll leave it up to you ;)

```js
this.children.find(child => child.innerText.trim()[0] === event.key)
```

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+          if (child.innerText.trim()[0] === event.key) {
```
surma: This is case sensitive. Maybe do `toLower()` on both?

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+            event.preventDefault();
+            child.focus();
+            break;
+          }
+        }
+      }
+    }
+
+    open() {
+      this._setTabindex(true);
```
surma: Unnamed boolean parameters are always a bit unintuitive to read. Maybe split into `setTabIndex()` and `unsetTabIndex()`?

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+         let prev = this._previousMenuItem(el) || this._lastMenuItem;
+         prev.focus();
+         break;
+       case KEYCODE.ESCAPE:
+         // If escape, close the menu.
+         event.preventDefault();
+         this.opened = false;
+         break;
+       case KEYCODE.TAB:
+         // If escape, close the menu.
```
surma: Comment talks about escape ;)

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+        }
+        prev = prev.previousElementSibling;
+      }
+      return null;
+    }
+
+    /**
+     * Handles keyboard interaction.
+     */
+    _onKeyDown(event) {
```
surma: `_onKeyDown` doesn’t handle Enter currently

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+    _upgradeProperty(prop) {
+      if (this.hasOwnProperty(prop) || this.hasAttribute(prop)) {
+        let value = this[prop];
+        delete this[prop];
+        this[prop] = value;
+      }
+    }
+
+    _onFocusOut(event) {
+      if (event.relatedTarget === null || !this.contains(event.relatedTarget)) {
```
surma: Woah, TIL `focusout`. 

This is super neat. Also cc @robdodson

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,213 @@
...
+    disconnectedCallback() {
+      this.removeEventListener('keydown', this._onKeyDown);
+      this.removeEventListener('focusout', this._onKeyDown);
+    }
+
+    /**
+     * Resets the menu by removing tabindex on all menu items.
+     */
+    _setTabindex(set) {
+      const children = Array.from(this.children);
```
surma: You can call `forEach` on `NodeList`, so `Array.from()` is not necessary. You could also do a `for...of` loop.

elements/howto-menu/howto-menu.js
devnook: I kinda agree. Hiding complexity inside a getter can lead to potential perf problems.

elements/howto-menu/howto-menu.js
devnook: Yes, in this version menuitems are not activateable (is this even a word?). Implementing submenus is a bigger deal so I'd rather do it once basic version lands.

elements/howto-menu/howto-menu.js
devnook: Hm, elegant, but less readable imo (which is arguable, depends on what you're used to I guess). Also would need to convert to array, so the gain is not worth it because perf.

elements/howto-menu/howto-menu.js
devnook: Hm, I think it should be case sensitive? There might be "other option" and "Other option" and they're not the same option.

elements/howto-menu/howto-menu.js
devnook: Funny thing, this.children does not have forEach, and this.childNodes does have:)

devnook:
PTAL

elements/howto-menu/howto-menu.js
surma: That’s fair!

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,220 @@
...
+        child = next.previousElementSibling;
+      }
+      return null;
+    }
+
+    /**
+     * Checks if a node is a 'menuitem', 'menuitemcheckbox' or 'menuitemradio'.
+     */
+    _isMenuItem(node) {
+      let ariaRoles = ['menuitem', 'menuitemcheckbox', 'menuitemradio'];
```
surma: Move this out to a constant along with `KEYCODES`

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,220 @@
...
+         // If tab, close the menu.
+         event.preventDefault();
+         this.opened = false;
+         break;
+       default:
+         break;
+      }
+      // If letter key, move to an item which starts with that letter.
+      if (event.keyCode >= KEYCODE.A && event.keyCode <= KEYCODE.Z) {
+        for (let i = 0, child; child = this.children[i]; i++) {
```
surma: `for...of`?

elements/howto-menu/demo.html
```diff
@@ -0,0 +1,37 @@
...
+     <span role="menuitem">first option</span>
+     <span role="menuitem">second option</span>
+     <span role="menuitem">other option</span>
+     <span role="menuitem">yet another option</span>
+   </howto-menu>
+</div>
+
+<script>
+var menu = document.querySelector('howto-menu');
+  document.getElementById('menu-btn').addEventListener('click', e => {
```
robdodson: Is this indentation off a bit?

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,220 @@
...
+    open() {
+      this._addTabindex();
+      let first = this._firstMenuItem();
+      first && first.focus();
+    }
+
+    set opened(isOpened) {
+      if (!!isOpened) {
+        this.setAttribute('opened', '');
+        this.open();
```
robdodson: Suggest moving into an attributeChangedCallback. This is one of the best practices recommended by the Custom Element spec author. Essentially all getters/setters just reflect to attribute and any side effects are handled in attributeChangedCallback

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,220 @@
...
+      } else {
+        this.removeAttribute('opened');
+        this._removeTabindex();
```
robdodson: Suggest moving to attributeChangedCallback

elements/howto-menu/howto-menu.js
robdodson: nice 👍 

that blur doesn't bubble can be pretty annoying

elements/howto-menu/howto-menu.js
devnook: Thanks! Was wondering how to divide responsibilities here, this makes it clear 👍

devnook:
PTAL

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,247 @@
...
+    open() {
+      this._addTabindex();
+      let first = this._firstMenuItem();
+      first && first.focus();
+    }
+
+    set opened(isOpened) {
+      if (!!isOpened) {
+        this.setAttribute('opened', 'true');
+        this.open();
```
robdodson: you can remove this because it's triggered in attributeChangedCallback now

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,247 @@
...
+      } else {
+        this.removeAttribute('opened');
+        this._removeTabindex();
```
robdodson: you can remove this one as well

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    open() {
+      this._addTabindex();
+      let first = this._firstMenuItem();
+      first && first.focus();
+    }
+
+    set opened(isOpened) {
+      if (!!isOpened) {
```
robdodson: small nit. would you mind replacing this with a call to `Boolean()`, similar to
https://github.com/GoogleChrome/howto-components/blob/master/elements/howto-checkbox/howto-checkbox.js#L125

My concern is not everyone understands what the !! trick does.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * `attributeChangedCallback` is called when any of the attributes in the
+     * `observedAttributes` array are changed. It's a good place to handle
+     * side effects, like changing the state of the widget.
+     */
+    attributeChangedCallback(name, oldValue, newValue) {
+      if (name === 'opened') {
+        if (!!newValue) {
```
robdodson: Would you mind switching to the pattern shown in howto-checkbox
https://github.com/GoogleChrome/howto-components/blob/master/elements/howto-checkbox/howto-checkbox.js#L154 ?

Your way is fine as well, I just want to be consistent throughout :)

devnook: Do you mean a switch on "name" variable? It seems kinda weird to switch in case of only 1 value ... Or switch on "newValue"? But then switch on a true/false also seems weird.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,245 @@
...
+          if (child.innerText.trim()[0] === event.key) {
+            event.preventDefault();
+            child.focus();
+            break;
+          }
+        }
+      }
+    }
+
+    open() {
```
robdodson: Would you mind making this a private method with an underscore? I think in general we've been trying to avoid exposing public methods as much as possible, especially when they mimic the functionality of properties. The reasoning is that it's much easier to pass state into properties and their corresponding attributes. So the element can be managed in a more declarative fashion, and it's easier for top-down data flow libraries to talk to them.

elements/howto-menu/howto-menu.js
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * `attributeChangedCallback` is called when any of the attributes in the
+     * `observedAttributes` array are changed. It's a good place to handle
+     * side effects, like changing the state of the widget.
+     */
+    attributeChangedCallback(name, oldValue, newValue) {
+      if (name === 'opened') {
+        if (Boolean(newValue)) {
```
robdodson: I think this is fine

robdodson:
@surma PTAL

surma:
🎉 🎉 🎉 🎉 🎉 🍾

robdodson:
WOOOOOO! Awesome work @devnook! Thank you 🦄

devnook:
YaY! I think it was the longest lasting Pr in my life (or three Prs):D

PR #99, opened 2017-08-04
Add checkbox to shadow dom

robdodson:
@surma I wanted to put this on your radar. Per the best practices doc I shared previously, I'm planning to move the checkbox styles into the shadow dom. I've come around to thinking that an element that attempts to style itself or its children will need a shadow root. Otherwise if someone tries to use that element inside of _another_ element that has a shadow root. it'll break.

Moving the styles into the element creates weirdness for the docs directory though. I had to remove all the image path variables. I'm not sure what the best way to work around that is?


surma:
LGTM. Merge at will

PR #100, opened 2017-08-11
Clean up howto-tabs

surma:
Probably best to review the entire code after all this time if I missed any patterns that we found in the meantime.

### DIFF
--- a/elements/howto-tabs/demo.html
+++ b/elements/howto-tabs/demo.html
@@ -11,40 +11,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <style>
-  howto-tabs {
-    /**
-     * The element uses flex box to line up the tabs in the first line
-     * and wraps if necessary.
-     */
-    display: flex;
-    flex-wrap: wrap;
-  }
   howto-tabs-tab {
     border: 1px solid black;
     padding: 20px;
   }
   howto-tabs-panel {
-    /**
-     * Each panel has a base size of 100%, forcing it to be in its
-     * own row.
-     */
-    flex-basis: 100%;
     padding: 20px;
     background-color: lightgray;
-  }
-  howto-tabs-panel[hidden] {
-    display: none;
   }
   howto-tabs-tab[selected] {
     background-color: bisque;
   }
 
   /**
-   * If JavaScript does not run, the element will stay `:unresolved`.
+   * If JavaScript does not run, the element will not match `:defined`.
    * In that case this style adds spacing between tabs and previous panel.
    */
-  howto-tabs:unresolved howto-tabs-tab:unresolved {
-    margin-top: 10px;
+  howto-tabs:not(:defined), howto-tabs-tab:not(:defined), howto-tabs-panel:not(:defined) {
+    display: block;
   }
 </style>
 
--- a/elements/howto-tabs/demo.html
+++ b/elements/howto-tabs/demo.html
@@ -11,15 +11,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <style>
-  howto-tabs-tab {
+  howto-tab {
     border: 1px solid black;
     padding: 20px;
   }
-  howto-tabs-panel {
+  howto-panel {
     padding: 20px;
     background-color: lightgray;
   }
-  howto-tabs-tab[selected] {
+  howto-tab[selected] {
     background-color: bisque;
   }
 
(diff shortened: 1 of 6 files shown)


robdodson:
One piece of feedback I got from Brian Kardell was to not do the role switcheroo. I go back and forth on that one...

elements/howto-tabs/demo.html
```diff
@@ -11,40 +11,24 @@
...
 <style>
-  howto-tabs {
-    /**
-     * The element uses flex box to line up the tabs in the first line
-     * and wraps if necessary.
-     */
-    display: flex;
-    flex-wrap: wrap;
-  }
   howto-tabs-tab {
```
robdodson: any reason you didn't want to put these styles in the shadow dom as well?

elements/howto-tabs/howto-tabs.js
```diff
@@ -27,13 +27,26 @@
...
   // To avoid invoking the parser with `.innerHTML` for every new instance, a
-  // template for the contents of the ShadowDOM is  is shared by all
+  // template for the contents of the ShadowDOM is shared by all
   // `<howto-tabs>` instances.
-  const shadowDOMTemplate = document.createElement('template');
-  shadowDOMTemplate.innerHTML = `
-    <slot name="tab"></slot>
-    <slot name="panel"></slot>
-  `;
+  const template = Object.assign(
```
robdodson: omg u such a hipster

robdodson: I feel like this is actually more code than the way I was doing it :D what's the upside of using object.assign?

elements/howto-tabs/howto-tabs.js
```diff
@@ -57,10 +70,15 @@
       this.attachShadow({mode: 'open'});
       // Import the shared template to create the slots for tabs and panels.
       this.shadowRoot.appendChild(
-        document.importNode(shadowDOMTemplate.content, true)
+        document.importNode(template.content, true)
```
robdodson: I was wondering recently if we have to keep using document.importNode cuz we're not pulling in anything from other documents

elements/howto-tabs/howto-tabs.js
```diff
@@ -78,12 +96,10 @@
...
-      // Currently, `slotchange` does not fire when an element is upgraded. For
-      // this reason, the element always processes the slots after the inner
-      // elements have been defined. If the current behavior of the `slotchange`
-      // event is change (as proposed in
-      // [this issue](https://github.com/whatwg/dom/issues/447)), the code below
-      // can be removed.
+      // Up until recently, `slotchange` events did not fire when an element is
+      // upgraded by the parser. For this reason, the element invokes the
+      // handler manually. Once the new behavior lands in all browsers, the code
+      // below can be removed.
```
robdodson: do you still want to link to the issue?

elements/howto-tabs/howto-tabs.js
```diff
@@ -78,12 +96,10 @@
...
       Promise.all([
         customElements.whenDefined('howto-tabs-tab'),
```
robdodson: wdyt about calling them <howto-tab> and <howto-panel> ?

elements/howto-tabs/demo.html
surma: Those are just the embellishments. The element could provide default styles, I guess, but I thought I’d only inline the functional styles (and also these need to be here when JS is not working).

elements/howto-tabs/howto-tabs.js
surma: It’s a pattern I have adopted. Probably not worth the confusion ;)

elements/howto-tabs/howto-tabs.js
surma: I was asking myself the same question. Not sure why I used it in the first place. I feel like it had a reason, but I don’t remember :-/

elements/howto-tabs/howto-tabs.js
surma: I removed the link because the spec change has been merged and it doesn’t say anything about the current state of individual browsers, which is probably what most readers would be interested in.

elements/howto-tabs/howto-tabs.js
surma: I don’t understand.

surma:
> One piece of feedback I got from Brian Kardell was to not do the role switcheroo. I go back and forth on that one...

But what’s the reasoning? I feel the PE is a pretty strong argument.

elements/howto-tabs/howto-tabs.js
devnook: Why is it sometimes window.customElements and sometimes just customElements?

surma: Who needs consistency???? (It should just be `customElements`).

elements/howto-tabs/howto-tabs.js
devnook: I would still consider linking to some doc on slochange (https://dom.spec.whatwg.org/#mutation-observers? or maybe if we have MutationObservers explained somewhere in our guide?) - it still feels like a relatively fresh concept and people might be confused whether it's a custom event or a standardised one (well, I was at first...).

elements/howto-tabs/howto-tabs.js
```diff
@@ -92,6 +108,15 @@
     }
 
     /**
+     * `disconnectedCallback` removes the event listeners that
+     * `connectedCallback` added.
+     */
+    disconnectedCallback() {
+      this.removeEventListener('keydown', this._onKeyDown);
+      this.removeEventListener('click', this._onClick);
```
devnook: How about touch events? Is it in scope?

elements/howto-tabs/howto-tabs.js
surma: Yeah, we should definitely link to MDN about MOs in our patterns section.

elements/howto-tabs/howto-tabs.js
surma: Good question. `click` triggers for touch events. Unless we want to handle touch interactions separately, I’d say we are good. But this is up for discussion.

surma:
PTAL

elements/howto-tabs/howto-tabs.js
robdodson: blargh sorry, `<howto-tabs>` and `<howto-panel>`

robdodson:
@surma I think Brian's point was that the ARIA docs say to not change the role on an element after it's hit the page. I guess the scenario I'm imagining where this would go wrong is if it takes a really long time for the js to load so someone navigates the page, and then suddenly all the elements are different and they wonder what happened

robdodson:
I think my only remaining open question is around whether we want to call the children `<howto-tab>` or `<howto-tabs-tab>`. I prefer `<howto-tab>` because it's just a bit nicer to read.

surma:
Fixed! Good to merge?

devnook:
Lgtm

robdodson:
awesome thanks man!

PR #104, opened 2017-08-16
Removing readme tips. These are on /web now

robdodson:


robdodson:
Going to self merge cuz in a bit of a hurry to get everything into /web.

PR #114, opened 2017-08-29
First stab at howto-label

robdodson:
work-in-progress, still need to add tests and tune up the comments. I think the implementation is ready for review though. @surma do you have time to give it a look? Also, what do you think is the best way to pull in another howto-component into my demo?

### DIFF
--- a/elements/howto-checkbox/howto-checkbox.js
+++ b/elements/howto-checkbox/howto-checkbox.js
@@ -30,7 +30,7 @@
     <style>
       :host {
         display: inline-block;
-        background: url('/elements/howto-checkbox/images/unchecked-checkbox.svg') no-repeat;
+        background: url('{%PATH%}/images/unchecked-checkbox.svg') no-repeat;
         background-size: contain;
         width: 24px;
         height: 24px;
@@ -39,17 +39,17 @@
         display: none;
       }
       :host([aria-checked="true"]) {
-        background: url('/elements/howto-checkbox/images/checked-checkbox.svg') no-repeat;
+        background: url('{%PATH%}/images/checked-checkbox.svg') no-repeat;
         background-size: contain;
       }
       :host([aria-disabled="true"]) {
         background:
-          url('/elements/howto-checkbox/images/unchecked-checkbox-disabled.svg') no-repeat;
+          url('{%PATH%}/images/unchecked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
       }
       :host([aria-checked="true"][aria-disabled="true"]) {
         background:
-          url('/elements/howto-checkbox/images/checked-checkbox-disabled.svg') no-repeat;
+          url('{%PATH%}/images/checked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
       }
     </style>
--- a/elements/howto-checkbox/howto-checkbox.js
+++ b/elements/howto-checkbox/howto-checkbox.js
@@ -30,7 +30,7 @@
     <style>
       :host {
         display: inline-block;
-        background: url('{%PATH%}/images/unchecked-checkbox.svg') no-repeat;
+        background: url('/elements/howto-checkbox/images/unchecked-checkbox.svg') no-repeat;
         background-size: contain;
         width: 24px;
         height: 24px;
@@ -39,17 +39,17 @@
         display: none;
       }
       :host([aria-checked="true"]) {
-        background: url('{%PATH%}/images/checked-checkbox.svg') no-repeat;
+        background: url('/elements/howto-checkbox/images/checked-checkbox.svg') no-repeat;
         background-size: contain;
       }
       :host([aria-disabled="true"]) {
         background:
-          url('{%PATH%}/images/unchecked-checkbox-disabled.svg') no-repeat;
+          url('/elements/howto-checkbox/images/unchecked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
       }
       :host([aria-checked="true"][aria-disabled="true"]) {
         background:
-          url('{%PATH%}/images/checked-checkbox-disabled.svg') no-repeat;
+          url('/elements/howto-checkbox/images/checked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
       }
     </style>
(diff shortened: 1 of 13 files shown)


elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+      const value = this.getAttribute('for');
+      return value === null ? '' : value;
+    }
+
+    set for(value) {
+      this.setAttribute('for', value);
+    }
+
+    attributeChangedCallback(name, oldVal, newVal) {
+      if (name === 'for') {
```
surma: Do we wanna do this even when only observing one attribute?

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+      return ['for'];
+    }
+
+    constructor() {
+      super();
+      this.attachShadow({mode: 'open'});
+      this.shadowRoot.appendChild(template.content.cloneNode(true));
+      // Listen for the slotchange event to label any wrapped children
+      this.shadowRoot.querySelector('slot')
+        .addEventListener('slotchange', this._forChanged.bind(this));
```
surma: It’s a bit weird to call `_forChanged` on `slotchange`. Maybe you can find a better name?

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+        target.removeAttribute('aria-labelledby');
+      }
+      target = this._findTarget();
+      target.setAttribute('aria-labelledby', this.id);
+    }
+
+    _onClick(event) {
+      let el = this._currentLabelTarget();
+      // If nothing is labeled, or if a wrapped child was clicked on, return.
+      if (!el || event.target === el) {
```
surma: Does this work with inner nested children? Like 

```html
<howto-label>
  <howto-checkbox></howto-checkbox>
  The <em>Label for this <strong>awesome</strong></em> checkbox
</howto-label>
```

I feel like it does because bubbling, but wanted to double-check.

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+     * If there is a for property, return the element with that id.
+     * Else, return the explicitly labeled child.
+     * Else, return the first element child (assume it's implicitly labeled).
+     */
+    _findTarget() {
+      if (this.for) {
+        // external target
+        let scope = this.getRootNode();
+        return scope.getElementById(this.for);
+      } else {
```
surma: I’d prefer no `else`

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+        let slottedChildren = this.shadowRoot
+          .querySelector('slot').assignedNodes({flatten: true});
```
surma: Worth caching the slot?

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,142 @@
...
+        let el;
+        slottedChildren.forEach(child => {
+          // Ignore text nodes
+          if (child.nodeType === Node.TEXT_NODE) {
+            return;
+          }
+          // Implicit internal target, first element child
+          if (!el) {
```
surma: This `if` is a bit confusing. Couldn’t you do it like this?

```js
// Get all non-text slotted children
let slottedChildren = 
  this._slot.assignedNodes({flatten: true})
    .filter(child => child.nodeType !== Node.TEXT_NODE);
// Find the first one that defines an explicit external target
let el =
  slottedChildren
    .find(child => child.hasAttribute('howto-label-target'));
// Return that first explicit external target, or the first child if there is none.
return el || slottedChildren[0];
``

**Comment by surma** [comment]
> Also, what do you think is the best way to pull in another howto-component into my demo?

Idk honestly. I’m wondering if we should make `<howto-label>` the exception and include it in every demo.

File: elements/howto-label/howto-label.js (hunk 3 above)
Comment (robdodson) [reply elements/howto-label/howto-label.js]: If you wrap additional element children then you have to put the `howto-label-target` attribute on the checkbox. Otherwise there's no way to know which element u want to be the label target.

File: elements/howto-label/howto-label.js (hunk 6 above)
Comment (robdodson) [reply elements/howto-label/howto-label.js]: yeah but now you're looping through the children twice :P

**Comment by robdodson** [comment]
leaving a note for myself that I need to style `:host` to use the default cursor instead of the text selection one. This is what `<label>` seems to do.

File: elements/howto-label/howto-label.js (hunk 3 above)
Comment (surma) [reply elements/howto-label/howto-label.js]: Understood.

File: elements/howto-checkbox/howto-checkbox.js [hunk 7]
```diff
@@ -30,7 +30,7 @@
     <style>
       :host {
         display: inline-block;
-        background: url('{%PATH%}/images/unchecked-checkbox.svg') no-repeat;
```
surma: Is this not working or why did you remove the placeholder?

elements/howto-label/howto-label.js
```diff
@@ -0,0 +1,151 @@
...
+      let slottedChildren = this._slot.assignedNodes({flatten: true});
+      let el;
+      slottedChildren.forEach(child => {
+        // Ignore text nodes
+        if (child.nodeType === Node.TEXT_NODE) {
+          return;
+        }
+        // Find the first element child. This is the implicit target
+        // of the label.
+        if (!el) {
```
surma: Still not very happy about this bit. I see your point but I think looping twice is a small price to pay for the readability here (and the number of children should be <10 on average, no?).

I won’t block on this, it just seems a bit weird (and harder to understand) to give the first iteration a special code branch.

elements/howto-checkbox/howto-checkbox.js
robdodson: oh I did it so I could use the element in my demo.html temporarily during testing. Still not sure what would be the best way to pull the checkbox into my demo.

elements/howto-label/howto-label.js
robdodson: Changing to your previously pitched approach to make the code easier to understand.

robdodson:
@surma PTAL. If you're cool with the e2e changes i can merge and we can take up doing a bigger e2e refactor down the line.

PR #115, opened 2017-09-02
howto-toggle-button

rajatsehgal:
An implementation of a howto-toggle-button in compliance with the wai-aria-practices as stated here: https://www.w3.org/TR/wai-aria-practices-1.1/#button
Followed similar pattern as howto-checkbox and added corresponding unit as well as e2e tests.

### DIFF
--- a/elements/howto-toggle-button/README.md
+++ b/elements/howto-toggle-button/README.md
@@ -1,0 +1,32 @@
+## Summary {: #summary }
+
+A `<howto-toggle-button>` represents a boolean option in a form. The most common type
+of toggle button is a dual-type which allows the user to toggle between two
+choices -- pressed and not pressed.
+
+The element attempts to self apply the attributes `role="button"` and
+`tabindex="0"` when it is first created. The `role` attribute helps assistive
+technology like a screen reader tell the user what kind of control this is.
+The `tabindex` attribute opts the element into the tab order, making it keyboard
+focusable and operable. To learn more about these two topics, check out
+[What can ARIA do?][what-aria] and [Using tabindex][using-tabindex].
+
+When the toggle button is pressed, it adds a `pressed` boolean attribute, and sets
+a corresponding `pressed` property to `true`. In addition, the element sets an
+`aria-pressed` attribute to either `"true"` or `"false"`, depending on its
+state. Clicking on the toggle button with a mouse, space bar or enter key, toggles these
+pressed states.
+
+The toggle button also supports a `disabled` state. If either the `disabled` property
+is set to true or the `disabled` attribute is applied, the toggle button sets
+`aria-disabled="true"` and set `tabindex="-1"`.
+
+## Reference {: #reference }
+
+- [Toggle Button Pattern in ARIA Authoring Practices 1.1][toggle-button-pattern]
+- [What can ARIA Do?][what-aria]
+- [Using tabindex][using-tabindex]
+
+[toggle-button-pattern]: https://www.w3.org/TR/wai-aria-practices-1.1/#button
+[what-aria]: https://developers.google.com/web/fundamentals/accessibility/semantics-aria/#what_can_aria_do
+[using-tabindex]: https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex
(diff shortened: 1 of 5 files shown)


elements/howto-toggle-button/howto-toggle-button.js
```diff
@@ -0,0 +1,233 @@
...
+
+  /**
+   * Cloning contents from a &lt;template&gt; element is more performant
+   * than using innerHTML because it avoids addtional HTML parse costs.
+   */
+  const template = document.createElement('template');
+  template.innerHTML = `
+    <style>
+      :host {
+        background-color: #eee;
```
surma: We try to keep visual styling out of the element. I’d say move stuff like `padding`, `background-color`, `border` etc to the demo file

elements/howto-toggle-button/howto-toggle-button.js
```diff
@@ -0,0 +1,233 @@
...
+        display: inline-block;
+        padding: 3px;
+        border: 1px solid #333;
+        border-radius: 3px;
+        cursor: default;
+        user-select: none;
+        transition: background-color .2s ease;
+      }
+      :host([hidden]) {
```
surma: I think `[hidden]` works out-of-the-box.

elements/howto-toggle-button/howto-toggle-button.js
rajatsehgal: I was a little confused about this one, I understand why. But, I saw that the howto-checkbox was doing it's styling through images and I needed similar basic styles for the button to work out of the box, I'll move them to the demo page instead. @robdodson any suggestions here?

elements/howto-toggle-button/howto-toggle-button.js
rajatsehgal: I was trying to follow the best practices here: https://developers.google.com/web/fundamentals/architecture/building-components/best-practices  "In addition to a default display style, add support for hidden with :host([hidden]) { display: none }". Also, howto-checkbox seems to do the same. But, I can remove it if that's the recommended practice.

elements/howto-toggle-button/howto-toggle-button.js
robdodson: I used images because they were sort of the easiest path to drawing a checkmark. But I've been planning to change that to something less opinionated. In general, we try to only use styles for the very basics of rendering the element correctly. This is partly to discourage folks from thinking this is a for-production element set :)

elements/howto-toggle-button/howto-toggle-button.js
robdodson: Yeah unfortunately `:host { display: inline-block }` will trump the UA stylesheet's `[hidden]` style. So Custom Elements need to explicitly add support for hidden back in. I think what you're doing here is correct.

elements/howto-toggle-button/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+When the toggle button is pressed, it adds a `pressed` boolean attribute, and sets
+a corresponding `pressed` property to `true`. In addition, the element sets an
+`aria-pressed` attribute to either `"true"` or `"false"`, depending on its
+state. Clicking on the toggle button with a mouse, space bar or enter key, toggles these
+pressed states.
+
+The toggle button also supports a `disabled` state. If either the `disabled` property
+is set to true or the `disabled` attribute is applied, the toggle button sets
+`aria-disabled="true"` and set `tabindex="-1"`.
```
robdodson: It looks like it removes `tabindex` when `disabled` is set. I think that behavior is fine, just wanna make sure to point that out in this comment.

robdodson:
This is awesome work @rajatsehgal! Thank you for putting this together 😃

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
woo! nice work @rajatsehgal 🐝 thanks again :)

robdodson:
@rajatsehgal check it out → https://www.youtube.com/watch?v=16gvkPfPIx4&index=1&list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g 😁

rajatsehgal:
@robdodson thanks for the shoutout! I really enjoy your videos. I'm planning to work on a howto-slider next, unless someone is already doing it?

robdodson:
Nobody is working on one right now so that would be great!

PR #119, opened 2017-09-25
Add initial aria-pressed state

robdodson:
Without an initial `aria-pressed` state, a screen reader will just announce that this is a "button", and not a toggle. Adding this initial state so it passes that hint on to the user.

@surma @rajatsehgal FYI


rajatsehgal:
👍

Issue #123, opened 2017-10-05
The howto-radio-group example doesn't have a label


slauriat:
The example has labels for each radiobutton, but not for the overall group. Looking at http://dash-elements.surma.link/0d0e37e38d2992bab729500d8efe9972d9f5501f/howto-radio-group/ I would expect a label of "Beverage" or something.

robdodson:
Good catch. I'll see if we can use [howto-label](https://github.com/GoogleChromeLabs/howto-components/tree/master/elements/howto-label) on it. That just recently landed.

robdodson:
Fixed with the power of fieldset

PR #124, opened 2017-10-07
Ratcheting

robdodson:
@surma FYI

Just tidying up some stuff. I did a little trick to make howto-label available in demos.

@slauriat I think this should address #123, can you verify I'm using fieldset correctly here:
https://github.com/GoogleChromeLabs/howto-components/compare/ratcheting?expand=1#diff-44055033e9c75ba1de2117d595729201R13

### DIFF
--- a/elements/howto-checkbox/howto-checkbox.js
+++ b/elements/howto-checkbox/howto-checkbox.js
@@ -38,16 +38,16 @@
       :host([hidden]) {
         display: none;
       }
-      :host([aria-checked="true"]) {
+      :host([checked]) {
         background: url('{%PATH%}/images/checked-checkbox.svg') no-repeat;
         background-size: contain;
       }
-      :host([aria-disabled="true"]) {
+      :host([disabled]) {
         background:
           url('{%PATH%}/images/unchecked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
       }
-      :host([aria-checked="true"][aria-disabled="true"]) {
+      :host([checked][disabled]) {
         background:
           url('{%PATH%}/images/checked-checkbox-disabled.svg') no-repeat;
         background-size: contain;
@@ -106,7 +106,7 @@
       this._upgradeProperty('checked');
       this._upgradeProperty('disabled');
 
-      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('keyup', this._onKeyUp);
       this.addEventListener('click', this._onClick);
     }
 
@@ -124,7 +124,7 @@
      * removing event listeners.
      */
     disconnectedCallback() {
-      this.removeEventListener('keydown', this._onKeyDown);
+      this.removeEventListener('keyup', this._onKeyUp);
       this.removeEventListener('click', this._onClick);
     }
 
@@ -190,7 +190,7 @@
       }
     }
 
-    _onKeyDown(event) {
+    _onKeyUp(event) {
       // Don’t handle modifier shortcuts typically used by assistive technology.
       if (event.altKey)
         return;
--- a/elements/howto-checkbox/README.md
+++ b/elements/howto-checkbox/README.md
@@ -19,7 +19,8 @@ checked states.
 
 The checkbox also supports a `disabled` state. If either the `disabled` property
 is set to true or the `disabled` attribute is applied, the checkbox sets
-`aria-disabled="true"` and set `tabindex="-1"`.
+`aria-disabled="true"`, removes the `tabindex` attribute, and returns focus
+to the document if the checkbox is the current `activeElement`.
 
 Warning: Just because you _can_ build a custom element checkbox, doesn't
 necessarily mean that you _should_. As this example shows, you will need to add
--- a/elements/howto-checkbox/demo.html
+++ b/elements/howto-checkbox/demo.html
@@ -15,7 +15,7 @@ limitations under the License.
   howto-checkbox {
     vertical-align: middle;
   }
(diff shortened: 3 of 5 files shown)


robdodson:
Oh I should also mention, I'm moving away from directly styling ARIA states, and instead treating ARIA as an implementation detail. Since setting `.checked` or `[checked]` on a `howto-checkbox` will also set `aria-checked` as a side effect, we can make it so the consumer never needs to know that ARIA is operating under the hood. Ultimately this will be good when something like AOM lands because Custom Elements wouldn't even need to sprout ARIA attributes, they could just do `this.acessibleNode.checked = true`.

elements/howto-checkbox/howto-checkbox.js
```diff
@@ -106,7 +106,7 @@
       this._upgradeProperty('checked');
       this._upgradeProperty('disabled');
 
-      this.addEventListener('keydown', this._onKeyDown);
+      this.addEventListener('keyup', this._onKeyUp);
```
surma: Is this intentional? I mean, for checkbox it makes sense to not rapidly change the `.checked` state when holding space, but for example in `howto-menu` should holding “arrow down” not cycle through the list?

surma (approved):
One minor Q, but LGTM :)

elements/howto-checkbox/howto-checkbox.js
robdodson: yeah that's why I limited it to just this component. I looked at what `<input type=checkbox>` does here and it doesn't rapidly toggle when you hold space. It _does_ show an active state which is something we could add as an extra layer of polish someday.

nathanhammond: It is also possible to trigger the down state by pressing on the associated label element so that a user can visually confirm the connection between label and checkbox. That is _why_ that state exists and isn't something that I consider to be polish. I'd love for this to actually be tackled since this library is intended to be a reference point.

A few additional notes while I'm here:
- It should disable dragging on the custom element.
- Global `mouseup` and (possibly) `keyup` listeners are required to reset the down state in case they don't occur on the same element.
- To get identical reading output to native and work around reading checkbox state backward in Chrome you'll need to use `aria-labelledby` to relate checkbox to label.
- Form submissions can be made to work by mirroring to a `display: none` `input type=checkbox`.
- Nested usage is a thing and is an especially good pattern for braille keyboards: `<label><checkbox> blah blah</label>`. Also, should work without having to specify identifiers.

/me disappears to my hole again.

robdodson: > It is also possible to trigger the down state by pressing on the associated label element so that a user can visually confirm the connection between label and checkbox. That is why that state exists and isn't something that I consider to be polish. I'd love for this to actually be tackled since this library is intended to be a reference point.

Interesting point about the label. The howto-label that was added in this PR does allow you to click on the label to toggle the control but neither support an `:active` state. I'll have to look into that as a future addition.

> It should disable dragging on the custom element.

Sorry for the dumb question, but I'm not sure I understand what you mean? Which CSS property are you referring to here? I don't think we allow any dragging but I'm probably missing something :)

> Global mouseup and (possibly) keyup listeners are required to reset the down state in case they don't occur on the same element.

I just checked the native `<input type=checkbox>` and it doesn't toggle state if you mousedown on the checkbox and then mouseup on a different node. It's as if you never clicked on it. That's the behavior we're emulating here. Again, maybe I'm missing something?

> Form submissions can be made to work by mirroring to a display: none input type=checkbox.

Yeah we mention this [in the docs](https://github.com/GoogleChromeLabs/howto-components/commit/7ca365291863f310c5006a01ec3029c62a485a6b#diff-10b8d4b4ed0ea3751111d373734e9fbaR24). I guess this is kind of an editorial decision where we didn't want to pick a specific solution because there's a few different ways folks could skin this cat. At TPAC we're going to see if we can get some folks to discuss an [onSubmitCallback](https://github.com/w3c/webcomponents/issues/187). I need to think about this one a bit more... I'm not opposed to injecting a hidden input because others have asked how to make a custom element work in a form. We would just need to call out that there are other ways to solve the problem (like creating a custom form element or submitting the form yourself using AJAX).

> Nested usage is a thing and is an especially good pattern for braille keyboards: <label><checkbox> blah blah</label>.

howto-label supports this.

> Also, should work without having to specify identifiers.

The only scenario where you need special identifiers for nested content is if you nest multiple elements within the label. That's mainly to avoid having to hard code a list of supported elements in our custom label and letting it support whatever custom element you want. I guess an alternative approach would be to have it find the first child with something like a `.value` property and consider that its target?

nathanhammond: > It should disable dragging on the custom element.

You can't click and drag a `input type=checkbox`. You can (in certain browsers, not sure which off the top of my head) drag a `div`. This is important for users with low motor accuracy. Click and accidentally drag and it shouldn't trigger this behavior. This is simply, `ondragstart` `event.preventDefault();`

> Global mouseup and (possibly) keyup listeners are required to reset the down state in case they don't occur on the same element.

This comment was regarding the down state only. If you "down" on the element and "up" on another node (or outside of the browser) you need to remove the down state. Attaching a listener to `document` (not `document.body`) handles this.

> Find the first child with something like a `.value` property and consider that its target

It's the set of aria roles in the [taxonomy](https://www.w3.org/TR/wai-aria/rdf_model.svg) for `input` plus the default form elements.

(Kinda doing this drive by, sorry for terseness.)

PR #129, opened 2017-10-28
<howto-slider>

sendilkumarn:

PR #131, opened 2017-11-21
<howto-slider> pt. 2

robdodson:
@sendilkumarn this PR addresses some changes from #129 

Because I wasn't able to push to that PR I had to create a new one.

The main changes I made were to turn ARIA into an implementation detail. So the element exposes `value, min, max` attributes and properties, and those reflect to ARIA but only as a side effect. Anyone consuming the component should never need to work with ARIA directly (the same way you don't have to work with ARIA when you're using `<input type="range>`).

I also tried to refactor the code so most things are driven off of setting the value property. So setting the `value` property triggers the `attributeChangedCallback` which in turn moves the `thumb`.

Also note that changing the `min` or `max` attributes/properties will set the value equal to itself again. This is useful if you have `value=50` and you change `min` from 0 to 50. The `thumb` will update and appear at the beginning of the track instead of in the middle.

I think this could use some more work but I think we're getting to a good place :)

### DIFF
--- a/elements/howto-slider/README.md
+++ b/elements/howto-slider/README.md
@@ -1,0 +1,36 @@
+## Summary {: #summary }
+
+A `<howto-slider>` represents a slider in a form. The slider allows user
+to select a value from a range of value.
+
+The element attempts to self apply the attributes `role="slider"` and
+`tabindex="0"` when it is first created. The `role` attribute helps assistive
+technology like a screen reader tell the user what kind of control this is.
+The `tabindex` attribute opts the element into the tab order, making it keyboard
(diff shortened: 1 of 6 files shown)


robdodson:
Oh also I removed the label from the shadow dom because `<input type="range">` doesn't have a built-in label. I realize they use one in the ARIA Authoring Practices example but I wanted to try to simplify the code as much as possible.

sendilkumarn:
@robdodson 👍 anything else has to be done over here

robdodson:
I think it's nearly done, sorry for the delay, the holidays got in the way. I'm back in the office this week and hopefully can wrap it up.

